PR #51, opened 2018-11-12
List Yarn as a dependency for building

saagarjha:
I'm trying to build this and realized it needs Yarn to build, which is not something that I have installed. I thought it might be useful to list this as a build requirement to help those like me who don't have it.

Anyways, onto figuring out how to tell deasync to target iOS instead of x86_64 through gyp…

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ You'll need these things to build the project:
 
  - Python 3
  - Ninja
+ - Yarn (only when building for iOS)
  - Meson (`pip install meson`)
  - Clang and LLD (on mac, `brew install llvm`, on linux, `sudo apt install clang lld` or `sudo pacman -S clang lld` or whatever)
 


README.md
```diff
@@ -16,6 +16,7 @@ You'll need these things to build the project:
 
  - Python 3
  - Ninja
+ - Yarn
```
tbodt: Yarn isn't required if you just want to run the emulator independently of the app for debugging.
```suggestion
 - Yarn (only when building for iOS)
```

saagarjha: Amended.

saagarjha:
I've never really done web development, so all of this stuff is new to me; I'm still trying to figure out how all of it works. I'm currently having an issue building some of the dependencies with Xcode, so I'll file an issue for that if I can't get that to work since it's somewhere deep inside the Yarn build.

PR #65, opened 2018-11-12
implement fcntl F_SETFL

lunixbochs:


tbodt:
The problem with this is the values of O_NONBLOCK and O_APPEND are different on Linux and Darwin. F_GETFL also has this problem.

lunixbochs:
Do you have a generalized structure for converting enums yet?

tbodt:
My handling of this issue in open() is here: https://github.com/tbodt/ish/blob/master/fs/real.c#L41

lunixbochs:
Are you open to having one big C file that's a bunch of functions like enum_in_mmap() enum_out_mmap() enum_in_open(), etc, with the same sort of pattern?

tbodt:
I would prefer to have functions like that closer to where they're used. For example, functions to map socket flags are defined in sock.h.

tbodt:
Merged along with flag mapping and a few other fixes.

PR #72, opened 2018-11-13
Add font size and theme picker

cmelbye:
**Changes**
- Add `UserPreferences` class which stores settings in `NSUserDefaults` to persist across app sessions.
- Add `ThemeViewController` to allow users to choose a theme. (currently supports Light and Dark themes)
- Update `AboutViewController` to save settings with `UserPreferences`.
- Update `Terminal` to evaluate JavaScript when `UserPreferences` are modified to change font color and size.
- Update `TerminalViewController` to set background color and status bar when `UserPreferences` are modified.

**To-Do**
- [ ] Terminal cursor is not updated. The `term.css` file uses a selector that I'm not sure how to update in JavaScript.
- [x] Keyboard and bottom bar are still white even when Dark theme is enabled. Bottom bar can easily be updated, but I haven't tested to see if `WKWebView` supports alternate keyboard styles.
- [ ] Caps lock setting is saved but doesn't actually modify the behavior of the caps lock key. I was not sure how to properly implement that logic.

**iOS Screenshots**

 

**iPad Screenshot**


Mnpn03:
This looks great to me! Is there a way to have the keyboard turn dark, too?

cmelbye:
@Mnpn03 I'm not sure how to do that yet but it sounds like it'll be possible after some trial and error. I'm not sure if an existing `WKWebView` will respect a new keyboard setting, or if it must be recreated after setting the new keyboard appearance. It's probably best to do this in a separate PR as this one has gotten quite large already.

tbodt:
Damn that's awesome 😄

A way to style the terminal that would work with the cursor would be to put all the user-configurable css in a <style> tag that's filled in by the javascript based on a template.

The keys in the extra row are the wrong color when using the dark theme. The color is an IBInspectable property on the buttons, not sure what the best way would be to keep that updated.

cmelbye:
Cool I can make those changes! Thanks

LoganDark:
Is there a slightly less high-contrast dark theme available? Pure `#ffffff` on pure `#000000` _will_ WRECK YOUR EYES.

tbodt:
You'll be able to choose your own colors for things, just as soon as we figure out how to make a reasonable color picker on iOS. (Or maybe have a text box where you can totally customize the stylesheet?)

LoganDark:
That could be extended to a custom CSS sort of thing, but then there comes the issue of a good text editor... (would totally be fine with a mini-terminal containing a good text editor though)

tbodt:
Blink implements themes as CSS stylesheets in a custom repo, along with a box where you can paste a URL. There are also a few nice themes included.

cmelbye:
I created a mockup for a possible solution of Custom Themes:

cmelbye:
It's also very easy to add a new enum case if you'd like to create a new preset theme. (Once this is merged.)

After adding the new enum case for your theme, the compiler will tell you where you need to update code for specifying colors.

tbodt:
That looks pretty nice! Status bar and keyboard should I think be a segmented control since there's only two possibilities for that; it might even be best to merge them (so you choose either dark keyboard + light status bar or vice versa).

tbodt:
Just updated the branch so it can be used with SQLite databases.

tbodt:
I think I got everything working...

PR #87, opened 2018-11-14
Add usage information instead of backtrace

lightlace:
When the user doesn't specify all of the arguments needed for `fakefsify.py` he get's a backtrace, which can be a little irritating.
Added a little usage information on how to use this script.


tbodt:
Thanks!

PR #88, opened 2018-11-14
Implement gettimeofday

saagarjha:
I wanted to poke around the internals of iSH, so I thought I'd "get my feet wet" by implementing a simple, obsolete, and mostly useless system call to get a taste of project development. Please let me know if there's anything you'd like changed; I've tried to model this after the other syscall implementations.

Just as a heads up, since this confused me for a while: if you want to test this, you'll need to use `syscall(SYS_gettimeofday, …)` directly instead of `gettimeofday` because musl's version just calls `clock_gettime`.

### DIFF
--- a/kernel/time.c
+++ b/kernel/time.c
@@ -5,6 +5,7 @@
 #include "debug.h"
 #include <time.h>
 #include <signal.h>
+#include <sys/time.h>
 #include "kernel/calls.h"
 #include "kernel/errno.h"
 #include "kernel/resource.h"
@@ -119,6 +120,25 @@ dword_t sys_times(addr_t tbuf) {
         tmp.tms_cstime = tmp.tms_stime;
         if (user_put(tbuf, tmp))
             return _EFAULT;
+    }
+    return 0;
+}
+
+dword_t sys_gettimeofday(addr_t tv, addr_t tz) {
+    STRACE("gettimeofday(0x%x, 0x%x)", tv, tz);
+    struct timeval timeval;
+    struct timezone timezone;
+    if (gettimeofday(&timeval, &timezone) < 0) {
+	    return errno_map();
+    }
+    struct timeval_ tv_;
+    struct timezone_ tz_;
+    tv_.sec = timeval.tv_sec;
+    tv_.usec = timeval.tv_usec;
+    tz_.minuteswest = timezone.tz_minuteswest;
+    tz_.dsttime = timezone.tz_dsttime;
+    if ((tv && user_put(tv, tv_)) || (tz && user_put(tz, tz_))) {
+	    return _EFAULT;
     }
     return 0;
 }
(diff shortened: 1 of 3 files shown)


lunixbochs:
fwiw I've run into real world code that invokes the `gettimeofday` syscall still, so not useless to have at all. It's also one of the functions in the vdso.

kernel/time.c
```diff
@@ -122,3 +123,22 @@ dword_t sys_times(addr_t tbuf) {
...
+    }
+    struct timeval_ tv_;
+    struct timezone_ tz_;
+    tv_.sec = timeval.tv_sec;
+    tv_.usec = timeval.tv_usec;
+    tz_.minuteswest = timezone.tz_minuteswest;
+    tz_.dsttime = timezone.tz_dsttime;
+    if ((tv && user_put(tv, tv_)) || (tz && user_put(tz, tz_))) {
+	    return _EFAULT;
+    }
```
tbodt: If tv is a valid pointer and tz is a nonzero invalid pointer, this will write out tv, when it should do nothing and return EFAULT. I don't think this is super important to fix though.

tbodt:
Implementation looks good. It sounds like it's been tested at least once, which is enough for this.

kernel/time.c
saagarjha: I took a moment to try out a variety of other Linuxes (Linuces?) do in this case, and I got different results, none of which actually match what you've mentioned. For systems with glibc, the program will SIGSEGV because it will [try to dereference `tz`](https://github.com/bminor/glibc/blob/master/sysdeps/posix/gettimeofday.c#L51), on others, where it actually performs the syscall, [`tv` will be set](https://github.com/torvalds/linux/blob/71e56028173bc84f01456a5679d8be9d681b49f1/kernel/time/time.c#L144). Honestly, I think we can consider behavior to be unspecified in this case, since you shouldn't have any expectation on what happens if you pass in non-`NULL` pointers that you shouldn't be passing in. I don't think POSIX specifies we do anything in this case.

tbodt: If linux has the same behavior, then it's definitely fine.

saagarjha:
Quick follow up question: do we want to have a `clock_settime`/`settimeofday` implementation as well? Obviously, it would have little effect on the system clock, but I was wondering if it was worth adding a stub that returns `EINVAL` just to make sure programs that depend on it don't exit out. Or maybe it's a feature that these programs die with an "missing syscall" instead of continuing in whatever inconsistent state they're left in?

tbodt:
I think it would be appropriate to stub those to just return EPERM. Who knows, there might be some program that always calls settimeofday and ignores the error code.

saagarjha:
> Who knows, there might be some program that always calls settimeofday and ignores the error code.

That was what I was scared about; programs that call `settimeofday` and assume that it worked and ending up borking something because they have no idea what the time really is. I guess we can just ignore those for now; I'll add the stubs and make a new pull request.

tbodt:
By the way @saagarjha are you in the Discord? If so can you tell me your name so I can give you your Contributor role?

saagarjha:
Just joined as @saagarjha.

PR #92, opened 2018-11-14
Implement setuid/setuid32

Ronsor:
This implements the `setuid` and `setuid32` system calls and adds permission checks to `setresgid`.


tbodt:
`setuid` is specified as changing the uid, euid, and suid. Also it should be allowed to set your euid to your uid or suid, even if neither are 0.

Ronsor:
setuid now has proper semantics

tbodt:
It should also allow setting your euid to your uid or suid, even if you're not root.

Ronsor:
currently `setreuid` are unimplemented; the proper behavior with setting the euid works fine in `setresuid`

tbodt:
A better implementation of setreuid would be to call setresuid with -1 as the last parameter. In fact, all the set*uid functions should probably just call setresuid.

tbodt:
setuid still isn't quite right, it doesn't allow a non-root process to set its uid to its euid. The check on setresuid is pretty broken, the correct behavior is specified here: http://man7.org/linux/man-pages/man2/setresuid.2.html#DESCRIPTION

tbodt:
setuid is now implemented, closing.

PR #95, opened 2018-11-15
Travis conf to build for iOS

xffzero:
its shiny
it works (maybe)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,6 @@
-matrix:
-    - os: linux
-    dist: trusty
-    sudo: required
-    language: c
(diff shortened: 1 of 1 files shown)


tbodt:
Can we get it to build both mac and linux?

xffzero:
i think we can build some stuff on linux, not the actual xcodeproj though

tbodt:
No, the app won't build on linux, but I'd like to keep doing linux builds like we are currently.

xffzero:
sounds good, i'll try and get it to work

tbodt:
Closing in favor of #98

PR #96, opened 2018-11-15
Make Xcode happy

saagarjha:
Use `(void)` instead of `()` for function prototypes, except for `fdtable_new` which actually takes an `unsigned`. Also, tell Xcode that this project uses spaces for indentation–if you don't, mine will keep trying to insert tabs everywhere because that's what my global configuration is set to.


tbodt:
I've already fixed the function prototypes, so I did a rebase so the first commit went away.

PR #105, opened 2018-11-17
fakefsify: handle hard links

zhuowei:
Some distributions such as Ubuntu contains hard links in the tar
file. Hard links share the same inode as a previously seen file.

### DIFF
--- a/tools/fakefsify.py
+++ b/tools/fakefsify.py
@@ -32,6 +32,8 @@ def extract_archive(archive, db):
             mode |= 0o060000
         elif member.isfifo():
             mode |= 0o010000
+        elif member.islnk():
+            pass
         else:
             raise ValueError('unrecognized tar entry type')
 
@@ -45,9 +47,15 @@ def extract_archive(archive, db):
             path.touch()
 
         inode = bytes(str(path.stat().st_ino), 'ascii')
+        if member.islnk():
+            # a hard link shares its target's inode
+            target_path = data/(member.linkname)
+            inode = bytes(str(target_path.stat().st_ino), 'ascii')
         meta_path = path.relative_to(data)
         meta_path = b'/' + bytes(meta_path) if meta_path.parts else b''
         db[b'inode ' + meta_path] = inode
+        if member.islnk():
+            continue
         db[b'stat ' + inode] = struct.pack(
             '=iiii',
             mode,


tbodt:
The correct way to handle hardlinks is to give them the same inode number in the database. The code in fs/fake-rebuild.c will then scan for those and create real hardlinks.

zhuowei:
@tbodt Thanks! I tried implementing hard links as you've suggested: now hard links are sharing one inode in the metadata:

```
>>> dbm.gnu.open("meta.db")
<_gdbm.gdbm object at 0x10c8d2f70>
>>> d = _
>>> d["inode /bin/uncompress"]
b'17239713'
>>> d["inode /bin/gunzip"]
b'17239713'
>>> d["stat 17239713"]
b'\xed\x81\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
```

I haven't tested if this works yet. I'll try to boot the result and make sure both targets of that link still work.

zhuowei:
Seems to work: the files share the same inode in ish.

stat_hard_link_compress

`chmod` doesn't work for me (always drops the last character in the filename) so can't test further.

tbodt:
Do you know why the space at the end of your prompt is missing?

zhuowei:
Not sure: and it's not the only place that drops the end character randomly. Probably the same reason

`chmod 644 /bin/gunzip`
prints
`chmod: cannot access '/bin/gunzi': No such file or directory`

and the STRACE debug also shows it tries to call stat with /bin/gunzi, even though the exec call had the correct name.

Maybe glibc's doing something weird with string formatting or something

tbodt:
It could be a CPU bug, you could try compiling with -Djit=false to see if that makes a difference.

zhuowei:
@tbodt Re. the missing space in the promot: A Linux x86_64 build with JIT enabled from current master branch (8bdd279c30ef7e985dfc7513a28203cf96bc0cd6) has the correct prompt, and `chmod 644 /bin/gunzip` works. So it seems the x86_64 JIT works.

Will create a new issue if I figure out what's causing it.

tbodt:
I just took the qemu test program and commented out everything that fails on x86_64, and I'm now running that on aarch64 to compare and see what's broken. I've already found and fixed an issue with operations on high byte registers (ah, bh, etc) in 
https://github.com/tbodt/ish/commit/3d102d8000126c547cf186c4219f9ffdaf7811de, and there's a good chance that that or some other fix I might push soon will fix it.

zhuowei:
@tbodt apt still doesn't work in Ubuntu since it needs Pentium Pro instructions (fcomi is the first instruction to fail), but otherwise bash, grep, whoami all work with a Ubuntu root image.

tbodt:
I implemented fcom as doing the same thing as fucom, since the difference only shows up when dealing with NaNs. You could do the same thing for fucomi.

PR #108, opened 2018-11-18
log: call NSLog instead of NSLogv

zhuowei:
Giving NSLogv a char* instead of a va_list crashes on iPad Air.
Switch to regular NSLog instead.

----

Why was NSLogv used here instead of NSLog? Is there something on macOS that prevents using regular NSLog?

### DIFF
--- a/kernel/log.c
+++ b/kernel/log.c
@@ -156,8 +156,8 @@ static void log_line(const char *line) {
 }
 #elif LOG_HANDLER_NSLOG
 static void log_line(const char *line) {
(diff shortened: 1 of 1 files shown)


tbodt:
I've already fixed this in 
https://github.com/tbodt/ish/commit/1c4bb920d9f9d2bf598577239a5da128b97c261d. NSLogv used to make sense, but I just redid the logging code and didn't realize that NSLogv didn't make sense with the new code until just now.

Issue #107, opened 2018-11-18
Illegal instruction in Ubuntu /bin/login when accessing TLS with lock cmpxchgl %ecx, %gs:-1879048060


zhuowei:
Trying to run [Ubuntu](
http://cdimage.ubuntu.com/ubuntu-base/releases/18.04.1/release/ubuntu-base-18.04.1-base-i386.tar.gz). After stubbing out sys_alarm to return _ENOSYS, am now running into this Illegal Instruction error:
```
NAME         PASS   STOP   NOTIFY
===========  =====  =====  ======
SIGUSR1      true   false  true 
2018-11-17 16:48:43.592477-0800 iSH[2803:382894] Could not load IOSurface for time string. Rendering locally instead.
2018-11-17 16:48:44.866403-0800 iSH[2803:382894] syncing
2018-11-17 16:48:44.915148-0800 iSH[2803:382894] [MC] System group container for systemgroup.com.apple.configurationprofiles path is /private/var/containers/Shared/SystemGroup/systemgroup.com.apple.configurationprofiles
2018-11-17 16:48:44.918140-0800 iSH[2803:382894] [MC] Reading from public effective user settings.
2018-11-17 16:48:44.968647-0800 iSH[2803:383045] 1 call 45  brk(0x0) = 0x56563000
2018-11-17 16:48:44.974194-0800 iSH[2803:383045] 1 call 122  = 0x0
2018-11-17 16:48:44.975558-0800 iSH[2803:383045] 1 call 33  faccessat(-100, "/etc/ld.so.nohwcap", 0x0, 0) = 0xfffffffe
2018-11-17 16:48:44.976555-0800 iSH[2803:383045] 1 call 33  faccessat(-100, "/etc/ld.so.preload", 0x4, 0) = 0xfffffffe
2018-11-17 16:48:44.977780-0800 iSH[2803:383045] 1 call 295 openat(-100, "/etc/ld.so.cache", 0x80000, 0x0) = 0x3
2018-11-17 16:48:44.978107-0800 iSH[2803:383045] 1 call 197 fstat64(3, 0xffffd6d0) = 0x0
2018-11-17 16:48:44.978373-0800 iSH[2803:383045] 1 call 192 mmap(0x0, 0x1bad, 0x1, 0x2, 3, 0) = 0xf7fcf000
2018-11-17 16:48:44.978501-0800 iSH[2803:383045] 1 call 6   close(3) = 0x0
[...]

PR #109, opened 2018-11-18
decode: handle GS prefix after LOCK prefix

zhuowei:
This fixes application startup on Ubuntu. #107

----

This, along with [a hack](https://github.com/zhuowei/ish/commit/b99dfb26c1a6afbfd401e05c045b8426e73c2e3a) to make all unimplemented syscalls return _ENOSYS, is enough to start Ubuntu's /bin/bash.

screen shot 2018-11-17 at 6 14 30 pm

Some questions:

1. I tried to model this after the regular GS handling code. Is the goto correct? It should restart decoding on the next prefix.
2. I think switching the order of these prefixes is [technically allowed](https://stackoverflow.com/questions/7197282/order-for-encoding-x86-instruction-prefix-bytes); are there any apps in the wild that has the GS prefix before LOCK?
3. Not related to this change, but : why do unimplemented syscalls raise a signal? Linux itself just [returns _ENOSYS](https://github.com/torvalds/linux/blob/master/kernel/sys_ni.c#L17). I understand that noting which apps use unimplemented syscalls is important, but shouldn't there be an option to ignore missing syscalls to get more apps to run?

### DIFF
--- a/emu/decode.h
+++ b/emu/decode.h
@@ -648,8 +648,10 @@ restart:
 
         // lock
         case 0xf0:
+            lockrestart:
             READINSN;
             switch (insn) {
+                case 0x65: TRACE("segment gs\n"); SEG_GS(); goto lockrestart;
 
 #define MAKE_OP_ATOMIC(x, OP, op) \
         case x+0x0: TRACEI("lock " op " reg8, modrm8"); \


tbodt:
1. The goto is correct.
2. I haven't seen anything that uses GS and LOCK in combination before. This seems strange to me actually, since GS is generally the base for thread-local storage, and synchronization shouldn't be needed for thread-local things.
3. An unimplemented syscall in Linux would be a system call number such as 222 that has no handler attached. If you make system call number 222, you get a SIGSYS.

zhuowei:
Thanks!

Issue #71, opened 2018-11-13
ffmpeg support


calebhicks:
Love what's happening here. Tried installing ffmpeg, and it worked! But all commands return Illegal instruction. 

My research leads me to believe it's a processor issue. Not sure what other information would be useful here, but happy to try.

Steps to reproduce:

```
apk add ffmpeg
ffmpeg
```

tbodt:
I've added to the list, though it's somewhat low priority.

calebhicks:
Fair enough! Thanks for all the work you're putting in to make it happen.

tbodt:
ffmpeg now segfaults instead of reporting "illegal instruction".

tbodt:
The mremap fix apparently fixed this as well as python.

cliss:
@tbodt Are you sure this is working? Perhaps it’s just my example, but I just tried and it was no dice.

I was going `mp4` → `mkv`. See the below.

```
Stream mapping:                                                                                                                                                   
  Stream #0:0 -> #0:0 (h264 (native) -> h264 (libx264))                                                                                                           
  Stream #0:1 -> #0:1 (aac (native) -> vorbis (libvorbis))                                                                                                        
Press [q] to stop, [?] for help                                                                                                                                   
Illegal instruction     
```

![image-1](https://user-images.githubusercontent.com/282460/49168362-4ef70e00-f305-11e8-8ce8-2ebd2c4f3ce6.jpg)

tbodt:
The fix isn't in the testflight build, did you build from source?

cliss:
Oh oh oh! Apologies; I *ass*umed this was in the most recent testflight build.

![](https://media.giphy.com/media/jUwpNzg9IcyrK/source.gif)

PR #111, opened 2018-11-18
decode: add fpu instructions needed by ffmpeg

zhuowei:
The list of added instructions:

- fsubr mem32
- fimuls mem16
- fidivrs mem16
- fdiv st(i), st
- fdivr st(i), st
- fpatan

fpatan's implementation calls down to normal atan2 since I don't know how to implement a 80-bit atan2.

This allows ffmpeg to start and convert 2 minutes of a .m4a file to WAV before segfaulting.

----

See #71.

I don't have testcases for the new instructions yet - how are you currently testing the FPU instructions?

### DIFF
--- a/emu/decode.h
+++ b/emu/decode.h
@@ -555,6 +555,7 @@ restart:
                     case 0xd80: TRACE("fadd mem32"); FADDM(mem_addr_real,32); break;
                     case 0xd81: TRACE("fmul mem32"); FMULM(mem_addr_real,32); break;
                     case 0xd84: TRACE("fsub mem32"); FSUBM(mem_addr_real,32); break;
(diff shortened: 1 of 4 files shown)


tbodt:
Implementations look good. The way I’ve been implementing transcendental functions is by using Python to prototype an algorithm to do it with just add/sub/mul/div, then translating that to C. But this is good enough for now, in fact it’s what qemu does for a lot of operations.

I don’t have any way to test fpu instructions currently beyond running ptraceomatic on a program that uses them. The qemu test program has some basic tests that are disabled right now, but they don’t at all cover the full space of fpu instructions. If you could find a program that tests every fpu instruction on a wide range of inputs, or write one, that would be great.

By the way, thank you for your contributions! Are you in the Discord server? If so, can I get your name so I can give you the Contributor role?

PR #114, opened 2018-11-18
decode: implement remaining xchg reg32, oax instructions

zhuowei:
This is just a copy-paste of the same code as the already implemented `xchg odi, oax` for the rest of the registers.


tbodt:
Thanks! Not sure why these weren't already implemented.

PR #116, opened 2018-11-19
Make copy/arrow keys/escape/a bunch of other stuff work again

saagarjha:
The current `-[TerminalView canPerformAction:withSender:]` breaks a bunch of things because it returns `NO` for a bunch of commands we can actually handle, such as copying or `handleKeyEquivalent:`.

### DIFF
--- a/app/TerminalView.m
+++ b/app/TerminalView.m
@@ -117,12 +117,6 @@ typedef enum {
     [self.terminal sendInput:data.bytes length:data.length];
 }
 
-- (BOOL)canPerformAction:(SEL)action withSender:(id)sender {
-    if (action == @selector(paste:))
-        return YES;
-    return NO;
-}
-
 - (void)paste:(id)sender {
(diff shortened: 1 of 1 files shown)


tbodt:
I was wondering why the hide keyboard button wasn't working. Does pasting from the context menu still work with this removed?

saagarjha:
Hang on, let me check. How exactly do you get the context menu to show up?

tbodt:
Long press on some text.

tbodt:
Checked the docs, apparently the default implementation of `canPerformAction:withSender:` just calls `respondsToSelector:`. If that's true then I'm smacking myself for not reading the docs earlier.

saagarjha:
…turns out having a breakpoint at `-[UIResponder canPerformAction:withSender:]` makes the thing not show up. D'oh. But yes, pasting from the context menu works.

tbodt:
Well then.

PR #118, opened 2018-11-19
decode: implement fxam

zhuowei:
There's no interpreter version yet.

----

This fixes the illegal instruction error when running `apt update` in a Ubuntu rootfs.

I haven't tested if this is accurate. `apt update` runs this instruction to show download progress. Since `select` with more than one fd doesn't work, it's stuck at 0%, so I can't tell if the instruction is broken.

### DIFF
--- a/emu/fpu.c
+++ b/emu/fpu.c
@@ -245,3 +245,27 @@ void fpu_patan(struct cpu_state *cpu) {
     ST(1) = f80_from_double(atan2(f80_to_double(ST(1)), f80_to_double(ST(0))));
     fpu_pop(cpu);
 }
+
+void fpu_xam(struct cpu_state *cpu) {
+    float80 f = ST(0);
+    int outflags = 0;
+    if (!f80_is_supported(f)) {
+        outflags = 0b000;
+    } else if (f80_isnan(f)) {
+        outflags = 0b001;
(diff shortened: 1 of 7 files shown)


emu/fpu.c
```diff
@@ -245,3 +245,27 @@ void fpu_patan(struct cpu_state *cpu) {
...
+        outflags = 0b001;
+    } else if (f80_isinf(f)) {
+        outflags = 0b011;
+    } else if (f80_iszero(f)) {
+        outflags = 0b100;
+    } else if (f80_isdenormal(f)) {
+        outflags = 0b110;
+    } else {
+        // normal.
+        // todo: empty
```
tbodt: Emptiness attribute for FPU stack registers is not implemented.

zhuowei: ah, ok.

tbodt:
Thanks!

zhuowei:
@tbodt This is [probably as far as I can go](https://github.com/zhuowei/ish/commits/master) for getting `apt` working in Ubuntu.

`apt update` runs `select` with multiple FDs, and I don't know how to implement that.
`apt show` shows "Unable to determine file size for fd 4 - fstat (2: No such file or directory)". Not sure why.

PR #178, opened 2018-12-10
reorganize device initialization for mem devs

lunixbochs:
I was going to implement /dev/random, but it seems like the best option is to proxy to the real file and I didn't want to dig into your fd mapping and flags stuff, so here's the cleanup I did to prep for it.

### DIFF
--- a/fs/dev.c
+++ b/fs/dev.c
@@ -1,6 +1,7 @@
 #include "kernel/errno.h"
 #include "fs/fd.h"
 #include "fs/dev.h"
+#include "fs/mem.h"
 #include "fs/tty.h"
 
 #pragma GCC diagnostic ignored "-Winitializer-overrides"
@@ -9,7 +10,7 @@ struct dev_ops *block_devs[] = {
 };
 struct dev_ops *char_devs[] = {
     [0 ... 255] = NULL,
-    [1] = &null_dev,
+    [1] = &mem_dev,
     [4] = &tty_dev,
 };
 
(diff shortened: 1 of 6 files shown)


tbodt:
I really don't think it's worth putting the null device in its own file, I'd want to put it in fs/mem.c. It also might be a good idea to make a new subdirectory for devices? Perhaps not yet.

tbodt:
The way I was planning on implementing /dev/u?random was writing a random_open function that would open /dev/u?random, put the resulting fd in real_fd, and install realfs_read/realfs_write/realfs_close.

lunixbochs:
Yes, but the open mode/flags should be preserved, or at least write mode + write write calls should be intercepted and null routed possibly with a permissions check (might not work on iOS anyway), stuff does write to it as root and it’s probably best to preserve the same error situations.  I mostly didn’t want to duplicate the big flag lookup from realfd_open.

lunixbochs:
Possible compromise: dev null/zero/full defined completely in their respective headers? Because they all need to go wherever null is, my problem was in dumping a ton of devices in one file.

Possible folder structure: fs/dev.c fs/dev/mem.c fs/proc/* and fs/sys/*?

tbodt:
I don’t think there’s anything wrong with putting a bunch of simple, related devices in the same file.

Maybe a better impl of u?random would be to have a global randomness generation function
that returns data from an eternally open /dev/urandom fd. Then anything that needs randomness (there are a couple places other than random devices) can use that.

lunixbochs:
The problem is consistency, there are 13 mem devices and some aren't as simple, but at that point IMO they should all or none go in mem.c.

I think get_random() is a totally reasonable API, that's how SecCopyRandomBytes on iOS works anyway. We should just proxy to either SecCopyRandomBytes or linux' getrandom(). Where would you put a global helper like that?

tbodt:
get_random would go in kernel/random.c I suppose.

PR #181, opened 2018-12-10
/dev/[u]random implementation

lunixbochs:
supercedes https://github.com/tbodt/ish/pull/178

reorganizes memory devices, adds device handlers for /dev/random, zero, full

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ macros:
     - &linux
       language: c
       os: linux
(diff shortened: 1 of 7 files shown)


tbodt:
There's still the slight problem of the /dev nodes not existing. Maybe those should be created on app launch if they don't exist?

lunixbochs:
I think that would be a good use of overlay mounts

PR #186, opened 2018-12-14
Allow fake ICMP raw sockets (ping working)

### LINKED ISSUE #76: ping does not work
Santiagos-iPad:~/stuff# ping google.com                                                                                                    
    PING google.com (216.58.193.142): 56 data bytes                                                                                            
    ping: can't create raw socket: Invalid argument

wallisch:
Requests for raw sockets for ICMP now create a DGRAM socket which allows ping to work.

Fixes #76

### DIFF
--- a/fs/sock.c
+++ b/fs/sock.c
@@ -312,13 +312,13 @@ dword_t sys_setsockopt(fd_t sock_fd, dword_t level, dword_t option, addr_t value
     char value[value_len];
     if (user_read(value_addr, value, value_len))
         return _EFAULT;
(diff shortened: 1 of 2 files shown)


tbodt:
Wow, didn't realize it would be that simple! Does traceroute work, or just ping?

tbodt:
Just tested, traceroute doesn't work, but ping does.

wallisch:
traceroute actually just needs SO_SNDBUF and IP_TTL translated. And unfortunately the recvmsg socket syscall, which is why I didn't include it in this PR ...

PR #206, opened 2018-12-19
Implement recvmsg socketcall (fixes traceroute)

### LINKED ISSUE #200: traceroute does not work
Hi, I've just installed iSH and tried to run `traceroute`. Unfortunately the program failed to run with following error:

    traceroute: setsockopt(SO_SNDBUF): Invalid argument

iSH Version: 1.0 (37)
iOS: 12.1.1
Device: iPhone SE

wallisch:
Implementation of the `recvmsg(3)` socketcall and some socket flags and sockopts for `traceroute`.

`sendmsg(3)` can be implemented the same way, the `msghdr` and child struct operations should then be moved to own functions.

`traceroute` freezes at some point for some hosts, though i don't know if this could be busybox's traceroute implementation.

Fixes #200

### DIFF
--- a/fs/sock.c
+++ b/fs/sock.c
@@ -369,6 +369,128 @@ dword_t sys_getsockopt(fd_t sock_fd, dword_t level, dword_t option, addr_t value
     return 0;
 }
 
+dword_t sys_recvmsg(fd_t sock_fd, addr_t msghdr_addr, dword_t flags) {
+    STRACE("recvmsg(%d, 0x%x, %d)", sock_fd, msghdr_addr, flags);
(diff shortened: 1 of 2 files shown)


tbodt:
Thanks! One question, why access fields in the various structures using offsets instead of defining structures?

wallisch:
Yeah i just noticed my stupidity, leave this open please, i will push an update with structs

tbodt:
I'm not too happy that this has to be over 100 lines of code, but there's no obvious way to make it less insane.

wallisch:
At least most of it can be reused for sendmsg(), recvmmsg() and sendmmsg()

PR #213, opened 2018-12-23
Upgrade Alpine to 3.8.2

0b101:


tbodt:
Not much to criticize here :)

PR #275, opened 2019-01-12
Added Green on Dark

ELChris414:
I have zero idea if this works as there is literally no way for me to test it. Please somebody with a mac and Xcode compile it and tell me if it works! Thanks

Suggestion from this by: sebastianpc#4420 on Discord

### DIFF
--- a/app/UserPreferences.m
+++ b/app/UserPreferences.m
@@ -148,16 +148,16 @@ static UIColor *UnarchiveColor(id data) {
 
 NSDictionary<NSString *, Theme *> *presetThemes;
 + (void)initialize {
-    presetThemes = @{@"Light":         [self _themeWithForegroundColor:UIColor.blackColor
+    presetThemes = @{@"Light": [self _themeWithForegroundColor:UIColor.blackColor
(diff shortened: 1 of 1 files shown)


app/UserPreferences.m
```diff
@@ -148,9 +148,11 @@ - (BOOL)isEqual:(id)object {
 
 NSDictionary<NSString *, Theme *> *presetThemes;
 + (void)initialize {
-    presetThemes = @{@"Light": [self _themeWithForegroundColor:UIColor.blackColor
+    presetThemes = @{@"Light":         [self _themeWithForegroundColor:UIColor.blackColor
```
tbodt: but the :s aren’t aligned anymore

tbodt: (in the method call I mean)

ELChris414: I aligned the [ instead:
image
anyways in the current file the [ are aligned

tbodt: weren’t they both aligned initially?

tbodt:
smooth

PR #280, opened 2019-01-16
 Updated AppDelegate.m so that IPv6 addresses are used for DNS

### LINKED ISSUE #173: Missing nameservers in /etc/resolv.conf
I don't seem to be able to get internet access with iSH without manually editing /etc/resolv.conf to add nameservers. A fresh install of iSH shows the following, where example is my university.

    # cat /etc/resolv.conf
    search example.ac.uk

This doesn't seem to work when the network is an 802.1X? I don't have access to another network to test, but I suspect this might be the awkward factor about the network I'm using that makes this less reproducible?

https://github.com/tbodt/ish/blob/7e44df935a6a7366be666a37331a024c8abfeaee/app/AppDelegate.m#L67-L75

If I do `echo nameserver 8.8.8.8 >> /etc/resolv.conf` then internet works fine and I can `apk update`.

DavidSouthgate:
Fixes #173 by allowing IPv6 DNS addresses

### DIFF
--- a/app/AppDelegate.m
+++ b/app/AppDelegate.m
@@ -71,6 +71,9 @@ static void ios_handle_exit(int code) {
     for (int i = 0; res.dnsrch[i] != NULL; i++) {
         [resolvConf appendFormat:@"search %s\n", res.dnsrch[i]];
     }
+    
+    int ns_count = 0;
+    
     for (int i = 0; i < res.nscount; i++) {
         if (res.nsaddr_list[i].sin_len == 0)
             continue;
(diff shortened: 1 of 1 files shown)


app/AppDelegate.m
```diff
@@ -65,21 +65,35 @@ - (int)startThings {
...
-    err = res_ninit(&res);
-    NSMutableString *resolvConf = [NSMutableString new];
-    for (int i = 0; res.dnsrch[i] != NULL; i++) {
-        [resolvConf appendFormat:@"search %s\n", res.dnsrch[i]];
+    res_state _state;
+    _state = malloc(sizeof(struct __res_state));
+    if (EXIT_SUCCESS != res_ninit(_state)) {
+        free(_state);
+        exit(2);
     }
```
tbodt: Why not include search domains anymore?

app/AppDelegate.m
```diff
@@ -65,21 +65,35 @@ - (int)startThings {
     exit_hook = ios_handle_exit;
     
     // configure dns
-    struct __res_state res;
-    err = res_ninit(&res);
-    NSMutableString *resolvConf = [NSMutableString new];
-    for (int i = 0; res.dnsrch[i] != NULL; i++) {
-        [resolvConf appendFormat:@"search %s\n", res.dnsrch[i]];
+    res_state _state;
+    _state = malloc(sizeof(struct __res_state));
```
tbodt: I don't think malloc/free is really necessary here, the res_state can still be stack allocated like it was before.

DavidSouthgate: Good point. Fixed

app/AppDelegate.m
DavidSouthgate: Honestly because I removed them accidentally. They have been added back again

PR #289, opened 2019-01-26
Add key commands for moving per word

kronawetter:
This maps Option + Right Arrow to move forward-word and Option + Left Arrow to move backward-word, like in the default key mapping of the macOS Terminal app.

I decided to create separate methods instead of extending `handleKeyCommand:` as this is not standard behavior of a shell. (similar to the existing Command + K key command for clearing the screen) If it is preferred to extend `handleKeyCommand:`, I am happy to update this PR accordingly.

### DIFF
--- a/app/TerminalView.m
+++ b/app/TerminalView.m
@@ -130,6 +130,14 @@
     return YES; // it's always ok to send a "delete"
 }
 
+- (void)backwardWord:(UIKeyCommand *)command {
+    [self insertText:@"\x1b" @"b"];
+}
+
+- (void)forwardWord:(UIKeyCommand *)command {
+    [self insertText:@"\x1b" @"f"];
+}
+
 - (void)clearScreen:(UIKeyCommand *)command {
     [self.terminal.webView evaluateJavaScript:@"term.clear()" completionHandler:nil];
 }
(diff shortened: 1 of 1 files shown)


kronawetter:
After submitting this PR, I started to realize that there might be some compatibility issues with these hard-coded sequences. I personally only move per words in the bash, where this sequence works just fine.

While I understand that using the Option key as Meta is a feature provided in most terminal applications, I do not like to use this setting as on non-English keyboards the Option key has to be used quite frequently.

Therefore, I tend to implement user-customizable key mappings. Does it currently make sense to do so or do you plan to introduce any changes soon that might require a rework of this functionality?

tbodt:
User-customizable key mappings would be awesome! And even if I was going to replace the terminal emulator tomorrow, the keyboard handling code would probably stay the same, so go right ahead.

kronawetter:
So, I’d like to keep this simple for now and similar to the functionality provided by the stock Terminal app on macOS.

Proposed differences from the Terminal app:
- Single profile only
- Only Cursor keys supported
  - For the first version, I'd like to base this on actions triggered by `UIKeyCommand`s only
    - [Only arrow and escape keys can be used as `UIKeyCommand` input](https://developer.apple.com/documentation/uikit/uikeycommand/input_strings_for_special_keys)
    - Num Pad = / * keys should also [be possible to support in theory](https://developer.apple.com/documentation/uikit/uikeymodifierflags/1621132-numericpad), but I don’t have hardware to test this
  - For a later version, some special key inputs can be derived from other sources
    - [Backward delete would be easy to support](https://developer.apple.com/documentation/uikit/uikeyinput/1614572-deletebackward), for example
- Only "Send text" action supported

Would you be okay with these features?

Screenshots of macOS Terminal app, for reference:
screen shot 2019-01-26 at 23 48 50
screen shot 2019-01-26 at 23 41 54
screen shot 2019-01-26 at 23 41 59

tbodt:
Closing as we now have option->meta mappings, which fulfills the original purpose of this PR.

PR #307, opened 2019-02-10
Upgrade Alpine to 3.9.0

0b101:
https://alpinelinux.org/posts/Alpine-3.9.0-released.html

### DIFF
--- a/iSH.xcodeproj/project.pbxproj
+++ b/iSH.xcodeproj/project.pbxproj
@@ -878,7 +878,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


tbodt:
Have you tested to make sure things are generally not broken? There were previously some issues with alpine edge where a lot of things like gcc would fail with "illegal instruction".

0b101:
During testing I did not see any issues with gcc. When installing nano (on the simulator) there were a couple errors related to ncurses unable to create and modify directories, but nano appears to function normally.

tbodt:
OK then, it's probably safe. The errors from installing nano are from the macOS filesystem being case insensitive. By a miraculous coincidence the iOS filesystem is case sensitive so everything works fine there.

saagarjha:
Recent versions of the simulator should simulate a case-sensitive filesystem, just FYI.

tbodt:
Interesting, when did they add that?

saagarjha:
[Mavericks][
https://worthdoingbadly.com/casesensitive-iossim/], it looks like.

PR #315, opened 2019-02-15
Added InterAppCommunication via /dev/iac

redbug26:
You have also to add TARGET_IS_EXTENSION in "Preprocessor Macros" like in this screen

sans titre 5

To test:
```
# Open iOS Settings
echo "App-prefs:root=General&path=Keyboard" > /dev/iac

# Open music
echo "music:" > /dev/iac

# Test x-callback (available also from any other application.. EG: type the url in safari)
echo "x-ish://x-callback-url/iac?output=username@gmail.com&password=myPassword" >/dev/iac
cat /dev/iac
```

Could be used to with Shortcut apps via something like
```
shortcuts://x-callback-url/run-shortcut?name=Calculate%20Tip&input=text&text=24.99&x-success=x-ish%3A%2F%2Fx-callback-url%2Fiac%3fresult
```

### DIFF
--- a/app/AppDelegate.m
+++ b/app/AppDelegate.m
@@ -5,6 +5,8 @@
 //  Created by Theodore Dubois on 10/17/17.
 //
 
+#import <Foundation/Foundation.h>
+
 #include <resolv.h>
 #include <arpa/inet.h>
 #include <netdb.h>
@@ -13,6 +15,9 @@
 #import "UserPreferences.h"
 #include "kernel/init.h"
 #include "kernel/calls.h"
+
+#import "IOSGateway.h"
+
 
 @interface AppDelegate ()
 
@@ -36,7 +41,7 @@ static void ios_handle_die(const char *msg) {
 
 - (int)startThings {
     NSFileManager *manager = [NSFileManager defaultManager];
-    NSURL *container = [manager containerURLForSecurityApplicationGroupIdentifier:@"group.app.ish.iSH"];
+    NSURL *container = [manager containerURLForSecurityApplicationGroupIdentifier:kGroupName];
     NSURL *alpineRoot = [container URLByAppendingPathComponent:@"roots/alpine"];
     [manager createDirectoryAtURL:[container URLByAppendingPathComponent:@"roots"]
       withIntermediateDirectories:YES
@@ -124,6 +129,9 @@ static void ios_handle_die(const char *msg) {
     generic_mknod("/dev/random", S_IFCHR|0666, dev_make(1, 8));
     generic_mknod("/dev/urandom", S_IFCHR|0666, dev_make(1, 9));
 
+    generic_mknod("/dev/iac", S_IFCHR|0666, dev_make(1, 99));
+
+    
     do_mount(&procfs, "proc", "/proc");
     do_mount(&devptsfs, "devpts", "/dev/pts");
 
@@ -134,6 +142,8 @@ static void ios_handle_die(const char *msg) {
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
     // get the network permissions popup to appear on chinese devices
     [[NSURLSession.sharedSession dataTaskWithURL:[NSURL URLWithString:@"http://captive.apple.com"]] resume];
+    
+    [[IOSGateway sharedSession] setup];
     
     [UserPreferences.shared addObserver:self forKeyPath:@"shouldDisableDimming" options:NSKeyValueObservingOptionInitial context:nil];
     int err = [self startThings];
@@ -146,6 +156,14 @@ static void ios_handle_die(const char *msg) {
         NSLog(@"failed with code %d", err);
     }
     return YES;
+}
+
+- (BOOL)application:(UIApplication *)application openURL:(NSURL *)url sourceApplication:(NSString *)sourceApplication annotation:(id)annotation
+{    
+    if ([[IOSGateway sharedSession] handleOpenURL:url]) {
+        return YES;
+    }
+    return NO;
 }
 
 - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary<NSKeyValueChangeKey,id> *)change context:(void *)context {
--- a/app/IOSGateway.h
+++ b/app/IOSGateway.h
@@ -1,0 +1,24 @@
+//
+//  IOSGateway.h
+//  iSH
+//
+//  Created by Miguel Vanhove on 15/02/2019.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface IOSGateway : NSObject
+
++ (IOSGateway *_Nullable ) sharedSession;
+
+- (void)setup;
+- (BOOL)handleOpenURL:(NSURL *)url;
+
+@end
+
+extern size_t iac_read(void *buf, size_t bufsize);
+extern size_t iac_write(const void *buf, size_t bufsize);
+
+NS_ASSUME_NONNULL_END
--- a/app/IOSGateway.m
+++ b/app/IOSGateway.m
@@ -1,0 +1,121 @@
+//
+//  IOSGateway.m
+//  iSH
+//
+//  Created by Miguel Vanhove on 15/02/2019.
+//
+
+#import "IOSGateway.h"
+#import "IACManager.h"
+#import "UIApplication+OpenURL.h"
+
+static IOSGateway *iOSGateway = nil;
+
+@interface IOSGateway ()
+{
+}
+
+@property (copy, nonatomic) NSData *iacResult;
+
+@end
+
+@implementation IOSGateway
+{
+}
+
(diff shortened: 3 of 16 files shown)


tbodt:
Thanks for taking the time to implement this! On first look this seems pretty good. One thing I'd want adjusted is the device number, which is currently 1,99. I would want to reserve an entire major number for iSH devices, and define an API for adding custom entries to the array in fs/dev.c. Looking at the list of allocated major numbers, 240 seems like a good choice, since major numbers from 240 to 254 are reserved for devices with no actual official major number. I'll review this a bit more later.

redbug26:
No problem :)
Working on /proc/dev/net now :)

app/AppDelegate.m
```diff
@@ -5,6 +5,8 @@
 //  Created by Theodore Dubois on 10/17/17.
 //
 
+#import <Foundation/Foundation.h>
```
saagarjha: Since this file imports AppDelegate.h, and by extension UIKit/UIKit.h, I don't think this is necessary?

app/IOSGateway.h
```diff
@@ -0,0 +1,24 @@
...
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface IOSGateway : NSObject
+
++ (IOSGateway *_Nullable ) sharedSession;
+
+- (void)setup;
+- (BOOL)handleOpenURL:(NSURL *)url;
```
saagarjha: I think this method would be better served being named `-(BOOL)canHandleOpeningURL:(NSURL *)url`.

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+//  Created by Miguel Vanhove on 15/02/2019.
+//
+
+#import "IOSGateway.h"
+#import "IACManager.h"
+#import "UIApplication+OpenURL.h"
+
+static IOSGateway *iOSGateway = nil;
+
+@interface IOSGateway ()
```
saagarjha: Is there a reason you're defining the interface for `IOSGateway` twice?

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+{
```
saagarjha: You don't need the brackets if they're empty, I don't think?

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+}
+
+@property (copy, nonatomic) NSData *iacResult;
+
+@end
+
+@implementation IOSGateway
+{
```
saagarjha: Ditto

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+}
+
++ (IOSGateway *_Nullable)sharedSession
```
saagarjha: Can this method return `NULL`? From what I see here, it doesn't look like it…

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+{
+    if (iOSGateway == nil) {
+        iOSGateway = [[IOSGateway alloc] init];
```
saagarjha: `[IOSGateway new]` also works, but this is a stylistic choice.

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    return iOSGateway;
+}
+
+- (id)init
+{
+    self = [super init];
+    if (self != nil) {
```
saagarjha: I like to be cute and stick the setting of `self` and check for `nil` in this condition, but again, this is a stylistic choice.

app/IOSGateway.m
```diff
@@ -0,0 +1,121 @@
...
+        
+        [UIApplication openURL:cmdString];
+    });
+
+#endif
+
+    return bufsize;
+}
+
+extern size_t iac_read(void *buf, size_t bufsize)
```
saagarjha: There's a bit of duplication here, perhaps there's a cleaner way to do this?

app/InterAppCommunication/IACClient.h
```diff
@@ -0,0 +1,50 @@
...
+/* Convenient methods to make call to external apps. If you create a subclass for your app, call these methods to launch the external app.
+*/
+- (void)performAction:(NSString*)action;
+
+- (void)performAction:(NSString*)action
+           parameters:(NSDictionary*)params;
+
+- (void)performAction:(NSString*)action
+           parameters:(NSDictionary*)params
+            onSuccess:(void(^)(NSDictionary*result))success
```
saagarjha: A space here would be nice

app/InterAppCommunication/IACManager.m
```diff
@@ -0,0 +1,260 @@
...
+- (BOOL)handleOpenURL:(NSURL*)url {
+    // An app can respond to multiple url schemes and the app can use different IACManagers for each one
+    // so we test if the url is handled by this manager
+    if (![url.scheme isEqualToString:self.callbackURLScheme]) {
+        return NO;
+    }
+    
+    // If the url is an x-callback-url compatible url we handle it
+    if ([url.host isEqualToString:kXCUHost]) {
+        NSString     *action     = [[url path] substringFromIndex:1];
```
saagarjha: There's a couple of places where you alternate between dot syntax and method syntax, such as here. It would be nice if you were consistent–personally, I lean towards dot syntax for most things that are "member-like" as this is (in my mind, this is usually amortized O(1) operations, or those that behave as such).

app/InterAppCommunication/NSString+IACExtensions.m
```diff
@@ -0,0 +1,71 @@
...
+
+#if !__has_feature(objc_arc)
+#error InterAppComutication must be built with ARC.
+// You can turn on ARC for only InterAppComutication files by adding -fobjc-arc to the build phase for each of its files.
+#endif
+
+
+@implementation NSString (IACExtensions)
+
++ (NSString*)stringWithUUID {
```
saagarjha: I'm not convinced this pulls its weight, especially since the alternative isn't particularly onerous (`[NSUUID UUID].UUIDString`)

app/InterAppCommunication/NSString+IACExtensions.m
```diff
@@ -0,0 +1,71 @@
...
+    CFUUIDRef uuid = CFUUIDCreate(kCFAllocatorDefault);
+    NSString *uuidStr = (__bridge_transfer NSString *)CFUUIDCreateString(kCFAllocatorDefault, uuid);
+    CFRelease(uuid);
+    
+    return uuidStr;
+}
+
+
+- (NSDictionary*)parseURLParams {
```
saagarjha: Perhaps [`NSURLComponents`](https://developer.apple.com/documentation/foundation/nsurlcomponents) might be of use?

app/AppDelegate.m
redbug26: true.. I remove it

app/IOSGateway.h
redbug26: why not ? More clear with the canHandleOpeningURL :)

app/IOSGateway.m
redbug26: True

app/IOSGateway.m
redbug26: It's to add the "NSData *iacResult" as private method

app/IOSGateway.m
redbug26: Stylistic ? :) I remove them

app/IOSGateway.m
redbug26: true

app/IOSGateway.m
redbug26: Are you java developer ? :) But Ok for the new :)

app/IOSGateway.m
redbug26: Yes.. It could. I like to be verbose (if running time is not impacted)
I clean the code

app/InterAppCommunication/IACManager.m
redbug26: I'm not the "InterAppCommunication" developer. It's an existing Cocoapods library by Antonio Cabezuelo Vivo

app/InterAppCommunication/NSString+IACExtensions.m
redbug26: I'm not the "InterAppCommunication" developer. It's an existing Cocoapods library by Antonio Cabezuelo Vivo

app/InterAppCommunication/NSString+IACExtensions.m
redbug26: I'm not the "InterAppCommunication" developer. It's an existing Cocoapods library by Antonio Cabezuelo Vivo

app/InterAppCommunication/IACClient.h
redbug26: I'm not the "InterAppCommunication" developer. It's an existing Cocoapods library by Antonio Cabezuelo Vivo

shzadakam (approved):

tajetaje:
What is needed to merge this?

PR #322, opened 2019-02-22
Improve memory management in JIT code

saagarjha:


tbodt:
Thanks, forgot about the jit hash being a separate allocation I guess...

PR #343, opened 2019-03-10
Add stub for stime

saagarjha:


tbodt:
What was using it?

saagarjha:
Nothing in musl or glibc; Android seems to to call it obsolete. I only added this because it was next to `ptrace` and easy to stub out–if someone *does* end up using it I think it's better to return `EPERM` rather than sending them `SIGSYS`.

tbodt:
There is no reason to use it when you could use settimeofday or clock_settime, but the stub is obvious enough that there's no reason not to merge this.

saagarjha:
Hence [this](https://github.com/torvalds/linux/blob/master/kernel/time/time.c#L95), [this](https://git.musl-libc.org/cgit/musl/tree/src/linux/stime.c#n8), and [this](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/stime.c;h=b0809be400e9706fe6780c147e78a58e1ea46849;hb=HEAD#l38).

PR #382, opened 2019-05-14
Add .editorconfig

MatthewMerrill:
See title. Uses 4 spaces and matches what I could figure out poking around in files.


tbodt:
Everything should have 4 space indent by default, not just .h

PR #384, opened 2019-05-15
remove status argument from halt_system

### LINKED ISSUE #383: halt_system mismatch in arg count
7a45420 introduces build error when I run `ninja` in build

```
../kernel/exit.c: In function ‘halt_system’:
../kernel/exit.c:124:1: error: number of arguments doesn’t match prototype
 static void halt_system() {
 ^~~~~~
../kernel/exit.c:9:13: error: prototype declaration
 static void halt_system(int status);
             ^~~~~~~~~~~
```

`status` argument [was removed from the halt_system's definition](https://github.com/tbodt/ish/commit/7a45420ac0bd67852f2a898a6f0d565f67d8cbf9#diff-2fab5c2f0efa9b89d9b17cc88d9b37fbL116) but [the prototype still had that argument](https://github.com/tbodt/ish/blob/7a45420ac0bd67852f2a898a6f0d565f67d8cbf9/kernel/exit.c#L9).

Fix could be to remove the argument from the prototype or add it back but ignored?

MatthewMerrill:
Fixes #383

### DIFF
--- a/kernel/exit.c
+++ b/kernel/exit.c
@@ -6,7 +6,7 @@
 #include "fs/fd.h"
 #include "fs/tty.h"
 
-static void halt_system();
+static void halt_system(void);
 
 static bool exit_tgroup(struct task *task) {
     struct tgroup *group = task->group;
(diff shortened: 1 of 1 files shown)


MatthewMerrill:
Fixes #383

MatthewMerrill:
updated

PR #387, opened 2019-05-18
add sys_call dup3 [330]

fishioon:


fs/fd.c
```diff
@@ -203,25 +203,36 @@ fd_t f_install(struct fd *fd, int flags) {
...
         return _EBADF;
     if (fd->inode != NULL) // temporary hack for files like sockets that right now don't have inodes but will eventually
         file_lock_remove_owned_by(fd, table);
     int err = fd_close(fd);
     table->files[f] = NULL;
-    bit_clear(f, table->cloexec);
+    if (flags & O_CLOEXEC_)
+    	bit_set(f, table->cloexec);
+    else
+    	bit_clear(f, table->cloexec);
```
tbodt: Setting a flag on a file descriptor that you just closed doesn't make sense. This should be set by the sys_dup3 function. That would also avoid having to add all this code to the close functions.

fs/fd.c
```diff
@@ -267,11 +278,16 @@ dword_t sys_dup2(fd_t f, fd_t new_f) {
...
-    f_close(new_f);
+    f_close_flags(new_f, flags);
     table->files[new_f] = fd;
     return new_f;
 }
 
+dword_t sys_dup2(fd_t f, fd_t new_f) {
+    return sys_dup3(f, new_f, 0);
+}
+
```
tbodt: nit: one blank line only please

fs/fd.c
fishioon: Thanks comment. I agree, but will get duplicate  code.
Can you add dup3? and let me learn more.

fishioon:
Neovim. dmesg show 'missing syscall 330'.
I am a new coder about kernel and please forgive me

fs/fd.c
tbodt: Can you explain why it would duplicate code?

tbodt:
Ah, I see. Neovim used to work fine but I guess something changed in an upgrade.

This is the right approach, thanks again for the PR.

tbodt:
@fishioon Are you planning on finishing this PR?

fishioon:
> @fishioon Are you planning on finishing this PR?

Sorry, I don't know how to finish it. Can you add 'sys_dup3' yourself and let me learn more?

PR #388, opened 2019-05-19
Upgrade Alpine to 3.9.4

0b101:
Change short list available at: https://alpinelinux.org/posts/Alpine-3.9.4-released.html


filfat:
Just tested and everything seems to be working fine (as it should since this is just a minor release).

PR #394, opened 2019-05-27
Added to the readme that the

joubin:
Added that Parcel is now required. I compiled and ran this app months ago and it work fine without it and now I need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ You'll need these things to build the project:
 
  - Python 3
  - Ninja
+ - Parcel `npm install -g parcel-bundler`
  - Yarn (only when building for iOS)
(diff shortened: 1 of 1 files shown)


tbodt:
You don't need that, it's supposed to be installed by the `npm install` in the script. But if node_modules already exists it won't run that, which can be fixed by re-running npm install if package.json changed.

joubin:
@tbodt, I've never had good luck with npm :( 

Which package.json do I run what on?

tbodt:
Xcode has a run script phase which runs npm install if node_modules doesn't exist. I'm changing that to run npm install if package.json is newer than node_modules.

PR #398, opened 2019-05-28
If stdin is not a tty, inherit stdin/stdout/stderr.

MatthewMerrill:
## Before:
```bash
$ echo "whoami" | ./ish -f ../alpine /bin/sh
whoami
/ # whoami
root
/ # ^[[20;5R
```
Program then hangs until interrupted.

## After:
```bash
$ echo "whoami" | ./ish -f ../alpine /bin/sh
root
$
```
Program stops!

This will be useful for automating running iSH for scripted tests.

### DIFF
--- a/kernel/init.c
+++ b/kernel/init.c
@@ -138,3 +138,26 @@ int create_stdio(const char *file) {
     current->files->files[2] = fd_retain(fd);
     return 0;
 }
+
+static struct fd *open_fd_from_actual_fd(int actual_fd) {
+    int fd_no = actual_fd;
+    if (fd_no < 0)
+        return ERR_PTR(errno_map());
+    struct fd *fd = fd_create(&realfs_fdops);
+    fd->real_fd = actual_fd;
+    fd->dir = NULL;
+    return fd;
+}
+
+int create_piped_stdio() {
+    struct fd *si = open_fd_from_actual_fd(STDIN_FILENO);
+    struct fd *so = open_fd_from_actual_fd(STDOUT_FILENO);
+    struct fd *se = open_fd_from_actual_fd(STDERR_FILENO);
+    si->refcount = 0;
+    so->refcount = 0;
+    se->refcount = 0;
+    current->files->files[0] = fd_retain(si);
+    current->files->files[1] = fd_retain(so);
+    current->files->files[2] = fd_retain(se);
+    return 0;
+}
--- a/kernel/init.c
+++ b/kernel/init.c
@@ -139,12 +139,11 @@ int create_stdio(const char *file) {
     return 0;
 }
 
-static struct fd *open_fd_from_actual_fd(int actual_fd) {
-    int fd_no = actual_fd;
+static struct fd *open_fd_from_actual_fd(int fd_no) {
     if (fd_no < 0)
         return ERR_PTR(errno_map());
     struct fd *fd = fd_create(&realfs_fdops);
-    fd->real_fd = actual_fd;
+    fd->real_fd = fd_no;
     fd->dir = NULL;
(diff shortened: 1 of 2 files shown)


xX_main_Xx.h
```diff
@@ -93,9 +93,14 @@ static inline int xX_main_Xx(int argc, char *const argv[], const char *envp) {
     if (err < 0)
         return err;
     tty_drivers[TTY_CONSOLE_MAJOR] = &real_tty_driver;
-    err = create_stdio(console);
-    if (err < 0)
-        return err;
+    if (isatty(fileno(stdin))) {
```
tbodt: s/fileno(stdin)/STDIN_FILENO

kernel/init.c
```diff
@@ -138,3 +138,25 @@ int create_stdio(const char *file) {
     current->files->files[2] = fd_retain(fd);
     return 0;
 }
+
+static struct fd *open_fd_from_actual_fd(int fd_no) {
+    if (fd_no < 0)
+        return ERR_PTR(errno_map());
+    struct fd *fd = fd_create(&realfs_fdops);
```
tbodt: Should use adhoc_fd_create, otherwise filesystem operations like fstat will crash.

Also this could return an error which should be propagated.

kernel/init.c
```diff
@@ -138,3 +138,25 @@ int create_stdio(const char *file) {
...
+    struct fd *si = open_fd_from_actual_fd(STDIN_FILENO);
+    struct fd *so = open_fd_from_actual_fd(STDOUT_FILENO);
+    struct fd *se = open_fd_from_actual_fd(STDERR_FILENO);
+    si->refcount = 0;
+    so->refcount = 0;
+    se->refcount = 0;
+    current->files->files[0] = fd_retain(si);
+    current->files->files[1] = fd_retain(so);
+    current->files->files[2] = fd_retain(se);
+    return 0;
```
tbodt: This would be simpler like `current->files->files[0] = open_fd_from_actual_fd(STDIN_FILENO)` etc. The refcount when the file is returned is just 1.

Also should check for if open_fd_from_actual_fd returns an error and pass it on.

xX_main_Xx.h
```diff
@@ -93,9 +93,14 @@ static inline int xX_main_Xx(int argc, char *const argv[], const char *envp) {
...
+        err = create_stdio(console);
+        if (err < 0)
+            return err;
+    }
+    else {
```
tbodt: join this line with the previous line please

kernel/init.c
```diff
@@ -138,3 +138,25 @@ int create_stdio(const char *file) {
     current->files->files[2] = fd_retain(fd);
     return 0;
 }
+
+static struct fd *open_fd_from_actual_fd(int fd_no) {
+    if (fd_no < 0)
+        return ERR_PTR(errno_map());
```
tbodt: This will never happen since you just pass in constants that are all >= 0

xX_main_Xx.h
```diff
@@ -93,9 +93,14 @@ static inline int xX_main_Xx(int argc, char *const argv[], const char *envp) {
...
+        create_piped_stdio();
```
tbodt: Handle error here

kernel/init.c
MatthewMerrill: How's this

tbodt: It's gonna need to use the IS_ERR family of macros.

tbodt: Update: it isn't, adhoc_fd_create just returns NULL instead of ERR_PTR(_ENOMEM).

PR #401, opened 2019-06-03
If stdout isn't a tty either, fall back to stdio

MatthewMerrill:
Should this check be done for stderr too?


tbodt:
No this is hopefully fine.

PR #404, opened 2019-06-07
float80 test should fail

MatthewMerrill:


tbodt:
Not quite "should fail" but more like "does fail and can't be assed to fix it".

PR #409, opened 2019-06-14
Updated the README

DavidSouthgate:


tbodt:
The common issues section is fine but installing parcel-bundler manually should not be required.

PR #411, opened 2019-06-18
WIP: Default theme based on iOS 13 Dark/Light mode

DavidSouthgate:
Added default theme which inherits system color scheme (i.e. Light mode and dark mode) when running iOS 13 or higher

### DIFF
--- a/app/UserPreferences.m
+++ b/app/UserPreferences.m
@@ -33,7 +33,14 @@ static NSString *const kPreferenceBootEnabledKey = @"Boot Enabled";
     self = [super init];
     if (self) {
         _defaults = [NSUserDefaults standardUserDefaults];
-        Theme *defaultTheme = [Theme presetThemeNamed:@"Light"];
+        
+        // Only iOS versions with dark mode have an system theme
+        Theme *defaultTheme;
+        if(@available(iOS 13.0, *))
+            defaultTheme = [Theme presetThemeNamed:@"System"];
+        else
+            defaultTheme = [Theme presetThemeNamed:@"Light"];
+        
         [_defaults registerDefaults:@{kPreferenceFontSizeKey: @(12),
                                       kPreferenceThemeKey: defaultTheme.properties,
                                       kPreferenceCapsLockMappingKey: @(CapsLockMapControl),
@@ -61,6 +68,9 @@ static NSString *const kPreferenceBootEnabledKey = @"Boot Enabled";
     [_defaults setObject:fontSize forKey:kPreferenceFontSizeKey];
 }
 
+- (NSNumber *)identifier {
+    return self.theme.identifier;
+}
 - (UIColor *)foregroundColor {
     return self.theme.foregroundColor;
 }
@@ -126,13 +136,16 @@ static UIColor *UnarchiveColor(id data) {
     if (self = [super init]) {
         _foregroundColor = UnarchiveColor(props[kThemeForegroundColor]);
         _backgroundColor = UnarchiveColor(props[kThemeBackgroundColor]);
+        _identifier = props[kThemeIdentifier];
     }
     return self;
 }
 
-+ (instancetype)_themeWithForegroundColor:(UIColor *)foreground backgroundColor:(UIColor *)background {
++ (instancetype)_themeWithForegroundColor:(UIColor *)foreground backgroundColor:(UIColor *)background identifier:(NSNumber *)identifier {
     return [[self alloc] initWithProperties:@{kThemeForegroundColor: ArchiveColor(foreground),
-                                              kThemeBackgroundColor: ArchiveColor(background)}];
+                                              kThemeBackgroundColor: ArchiveColor(background),
+                                              kThemeIdentifier: identifier
+                                              }];
 }
 
 - (UIStatusBarStyle)statusBarStyle {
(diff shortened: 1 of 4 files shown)


app/UserPreferences.m
```diff
@@ -166,29 +182,43 @@ - (BOOL)isEqual:(id)object {
...
+                                                backgroundColor:UIColor.blackColor
+                                                     identifier:@3]};
 }
 
 + (NSArray<NSString *> *)presetNames {
-    return @[@"Light", @"Dark", @"1337"];
+    NSArray* names = @[];
+    
+    // Only iOS versions with dark mode have an Auto theme
+    if(SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"13.0"))
```
tbodt: Use `if (@available(iOS 13.0, *))`

app/UserPreferences.m
```diff
@@ -166,29 +182,43 @@ - (BOOL)isEqual:(id)object {
...
+    
+    return [names arrayByAddingObjectsFromArray:@[@"Light", @"Dark", @"1337"]];
 }
 + (instancetype)presetThemeNamed:(NSString *)name {
     return presetThemes[name];
 }
 - (NSString *)presetName {
     for (NSString *name in presetThemes) {
-        if ([self isEqual:presetThemes[name]])
+        if ([self isEqual:presetThemes[name]]) {
```
tbodt: Could you leave this formatting alone? You're not changing this part of the file, so this commit would show up unnecessarily in git blame.

app/UserPreferences.m
```diff
@@ -154,8 +165,13 @@ - (UIKeyboardAppearance)keyboardAppearance {
 }
 
 - (NSDictionary<NSString *,id> *)properties {
+    NSLog(@"%@", self.identifier);
+    NSLog(@"%@", self.foregroundColor);
+    NSLog(@"%@", self.backgroundColor);
+    NSLog(@"");
```
tbodt: I assume these are left in by mistake?

app/UserPreferences.m
```diff
@@ -7,6 +7,7 @@
 
 #import <UIKit/UIKit.h>
 #import "UserPreferences.h"
+#define SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(v)  ([[[UIDevice currentDevice] systemVersion] compare:v options:NSNumericSearch] != NSOrderedAscending)
```
tbodt: Delete (see below regarding the if statement)

app/UserPreferences.m
```diff
@@ -61,6 +66,9 @@ - (void)setFontSize:(NSNumber *)fontSize {
     [_defaults setObject:fontSize forKey:kPreferenceFontSizeKey];
 }
 
+- (NSNumber *)identifier {
+    return self.theme.identifier;
+}
```
tbodt: Introducing a new identifier user defaults key that then has to live forever is a bad idea, in particular because I eventually want users to be able to create their own themes.

app/UserPreferences.m
DavidSouthgate: Done

app/UserPreferences.m
DavidSouthgate: Done

app/UserPreferences.m
DavidSouthgate: Yeah, they have been removed

app/UserPreferences.m
DavidSouthgate: In my latest commit I have changed these to just use the theme name. Do you think that's any better? Before the app was just using the background and foreground colours to identify each theme (so if two themes had the same colour scheme they would appear to be the same).

DavidSouthgate:
I have rebased the commit out. However, I have been unable to test what happens if the system coolers are changed from control centre (because the app doesn't currently work on beta 4).

DavidSouthgate:
Will update PR to ensure the theme is updated when you switch the system colors from control center, without restarting the app.

Regularcheckm8:
I’m going to reintroduce my pr for custom theming in a bit; if we could have this as an option... yeah idk. I’ve been thinking about this for a bit and if the dark mode is on, shouldn’t it dampen the colours? It gets confusing but when This gets pushed I’ll definitely have to do more work

kkk669:
@DavidSouthgate I also recommend adding the following changes to `LaunchScreen.storyboard`. `systemBackgroundColor` will follow the system theme.

```diff
diff --git a/app/Base.lproj/LaunchScreen.storyboard b/app/Base.lproj/LaunchScreen.storyboard
index f83f6fd..797c505 100644
--- a/app/Base.lproj/LaunchScreen.storyboard
+++ b/app/Base.lproj/LaunchScreen.storyboard
@@ -13,7 +13,7 @@
                     <view key="view" contentMode="scaleToFill" id="Ze5-6b-2t3">
                         <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
-                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
                         <viewLayoutGuide key="safeArea" id="6Tk-OE-BBY"/>
                     </view>
                 </viewController>
```

hazcod:
Would love a dark theme!

Mnpn03:
@hazcod It already exists under Settings > Appearance in iSH, this PR is just for automatic switching based on the iOS setting, but perhaps that's what you meant!

DavidSouthgate:
I don’t have enough time to complete this implementation so closing PR. Someone else please feel free to pick this up if you want.

upintheairsheep:
Now have time?

saagarjha:
This has already been implemented in master.

PR #418, opened 2019-06-27
Upgrade Alpine to 3.10.0

0b101:
https://alpinelinux.org/posts/Alpine-3.10.0-released.html


tbodt:
Could you also update https://github.com/tbodt/ish/blob/master/tests/e2e/e2e.bash#L11?

PR #426, opened 2019-07-15
movaps, movss, movsd, ucomiss, pxor, xorps

MatthewMerrill:
ucomiss will get more coverage with e2e qemu tests


brandonros:
what other SSE2 instructions are left before epoll works?

tbodt:
epoll has been working for a long time.

PR #431, opened 2019-07-18
Changed webView for iOS13b4

lorenzodla:
Changed the request for webView iOS13b4. Now it shows the interface again.


tbodt:
Thanks! Do you know why this stopped working?

lorenzodla:
According to the Xcode console, when you are doing the NSURLRequest for loading the html file, this file is outside the Sandbox. Also it says that the URL is not expected. It looks to me more than a bug than other thing. I leave here what it says.

`Received an unexpected URL from the web process: 'file:///var/containers/Bundle/Application/E98DA168-A6D8-4810-B417-048034ACC409/iSH.app/xterm-dist/term.html'`

`[Process] 0x10108c418 - WebPageProxy::Ignoring request to load this main resource because it is outside the sandbox`

About it working in other versions, it works. I have tested it on an iPhone with iOS 12.2 and it behaves the same than with the other method.

PR #441, opened 2019-07-22
Fix FUNDING.yml PayPal link

alexismarquis:
'paypal' is not a valid funding platform, so I guess it should just be a custom link.


tbodt:
Welp

PR #447, opened 2019-07-25
dyn_dev, devices.h and /dev/clipboard device (fix #192)

### LINKED ISSUE #192: Feature request: pbcopy/pbpaste style clipboard integration
On the Mac, you can use the `pbcopy`/`pbpaste` commands to copy from and paste to the clipboard. It would be awesome if you could do something similar in iSH.

The way I see this working is through virtual devices, but maybe there's a more clever way to do it. On iOS, `UIPasteboard` stores clipboard entries as pairs of UTI (Universal Type Identifier) and content. So the ideal API is probably more complicated than simply reading and writing to a single virtual device.

Right now iSH supports copy and paste on the terminal, but that isn't as powerful as you're limited to text. The only way to get data in and out of iSH is currently the file system, which is less convenient in some cases.

I think the feature in itself is pretty cool, but it would also be very powerful to use together with Siri Shortcuts (#59).

stek29:


fs/dyndev.h
```diff
@@ -0,0 +1,31 @@
+#ifndef FS_DYN_DEV_H
+#define FS_DYN_DEV_H
+
+// dyn_dev's are dynamically added devices (character only for now)
+// with custom dev_ops assigned
+// It's useful to add new device "drivers" in runtime (for example,
+// devices only present on some platforms)
+
+// (int)('D'+'Y') == 157
+#define DYN_DEV_MAJOR 157
```
tbodt: On Linux, major number 157 is reserved for a Specialix RIO serial card. Major numbers 234-254 are reserved for dynamic assignment, so we should use one of those.

fs/dyndev.h
```diff
@@ -0,0 +1,31 @@
...
+// dev_ops handing DYN_DEV_MAJOR major number
+extern struct dev_ops dyn_dev;
+
+// Registeres new device with major number DYN_DEV_MAJOR and provided
+// minor number handled by provided ops.
+//
+// ops should be valid for "kernel" lifetime (should not be freed, but
+// might be static), and should not be null
+//
+// minor should be 0-255 or -1. if it is -1, minor number is allocated
```
tbodt: Let's require each new device to choose a fixed minor number to avoid problems with a device file not working anymore. We can have a central header file to manage the allocation (similar to linux's devices.txt).

app/Pasteboard.m
```diff
@@ -0,0 +1,258 @@
...
+        free(fd_priv(fd).wbuffer);
+    }
+    return 0;
+}
+
+static int clipboard_open(int major, int minor, struct fd *fd) {
+    CLIP_DEBUG_PRINTK("%s(%p)\n", __func__, fd);
+
+    // Zero fd_priv data
+    bzero(&fd_priv(fd), sizeof(fd_priv(fd)));
```
tbodt: The man page for bzero says it's deprecated, use memset instead.

app/Pasteboard.m
```diff
@@ -0,0 +1,258 @@
+#include <string.h>
+#import <UIKit/UIKit.h>
+#include "fs/poll.h"
+#include "fs/dyndev.h"
+#include "kernel/errno.h"
+#include "debug.h"
+
+// Prepare for fd separation
```
tbodt: Message to future self: this is explained in discord.

app/Pasteboard.m
```diff
@@ -0,0 +1,258 @@
...
+#define fd_priv(fd) fd->clipboard
+
+#if 1
+#define CLIP_DEBUG_PRINTK(fmt, ...) printk(fmt, ##__VA_ARGS__)
```
tbodt: I'd prefer for debug logging which is mainly useful during writing the code to be removed before merging. If you'd rather not do that, use TRACE from debug.h, which can already be toggled with a compile time flag.

app/Pasteboard.m
```diff
@@ -0,0 +1,258 @@
...
+    CLIP_DEBUG_PRINTK("%s offset=%d pb_gen=%d len=%lu\n", tag, fd->offset, fd_priv(fd).generation, UIPasteboard.generalPasteboard.string.length);
+}
+
+
+// If pasteboard contents were changed since file was opened,
+// all read operations on in return error
+static int check_read_generation(struct fd* fd) {
+    UIPasteboard *pb = UIPasteboard.generalPasteboard;
+
+    uint64_t pb_gen = (uint64_t)pb.changeCount;
```
tbodt: I prefer putting a space after a cast, and putting a space before the * in a pointer type. This is somewhat my fault for not getting clang-format set up though.

stek29: @tbodt agreed on space before `*`, noted about space after cast.

app/Pasteboard.m
stek29: @tbodt Yup, I was planning to remove it once I was completely sure it's finished. Shouldn't have removed `WIP` tag yet :)

app/Pasteboard.m
stek29: re-posting it here:

What do you think about refactoring fd system a bit so fd data for different types of fd isn't placed in one union in one file, but instead is:
- replaced by fsprivate pointer (additional allocation for each fd, additional dereference for many operations), OR
- is placed after shared fd (I like this more, but it would require more refactoring i think)

For the second, additional field (fd_priv_size or smth) can be added to fd_opts, and fd_create would do malloc(sizeof(struct fd) + ops.fd_priv_size) instead of malloc(sizeof(struct fd)).

It'd be a great addition for dyn_dev, since right now I'd have to add clipboard related data into union in fd.h, but it really shouldn't be there.

fs/dyndev.h
stek29: Should I remove dynamic minor number selection to simplify dyn_dev_register?

fs/dyndev.h
stek29: Linux assigns them starting from 254 and downward -- should I go upward or downward (234 or 254)?

fs/dyndev.h
tbodt: Yeah that's what I was thinking.

app/Pasteboard.m
tbodt: FYI GitHub lets you mark a PR as "draft", which shows it in grey and prevents it from being merged

stek29:
Please don't merge, I'll refactor stuff a bit more (add devices.h)

fs/dyndev.h
tbodt: Doesn't really matter

stek29:
Oops, wsync is broken a bit, will fix tomorrow:

- write “hel”
- fsync
- write “lo”
- close/fsync

Expected: hello
Got: lo

stek29:
Ok, I think it's ready.

fs/fd.h
```diff
@@ -55,6 +55,14 @@ struct fd {
             // locked by fd->lock
             struct list unix_scm;
         } socket;
+
+        struct {
+            uint64_t generation;
+            void* wbuffer;
+            size_t wbuffer_off;
```
tbodt: Why is this separate from fd->offset?

fs/fd.h
```diff
@@ -55,6 +55,14 @@ struct fd {
...
+            size_t wbuffer_len;
+            size_t wbuffer_size;
```
tbodt: What's the difference between "len" and "size?

fs/fd.h
```diff
@@ -55,6 +55,14 @@ struct fd {
             // locked by fd->lock
             struct list unix_scm;
         } socket;
+
+        struct {
+            uint64_t generation;
+            void* wbuffer;
```
tbodt: What does "w" mean? What's in wbuffer?

stek29: I should really document this better

stek29:
So, here's how it works:

There's a buffer (wbuffer) of dynamic size `wbuffer_size`.
All `write`s go to it. `wbuffer_len` has length of data currently in buffer.

`wbuffer` is then "flushed" to UIPasteboard. When that happens, all data in `wbuffer` is copied to it, and `wbuffer_off` is adjusted -- it's the length of data currently in UIPasteboard -- aka start offset of `wbuffer`.

I'm not sure that `wbuffer_off` and partial flushes were worth it -- code complexity grew a lot, one case (write, flush, seek 0, read) no longer works, and all of that is to save just a bit of memory.


@tbodt check on commit 48008e8f06b0bde79f885c7b16e41a9d4e527ac6 -- it should be simpler (and more easily understandable)

fs/fd.h
stek29: see https://github.com/tbodt/ish/pull/447#issuecomment-516057341

fs/fd.h
stek29: See https://github.com/tbodt/ish/pull/447#issuecomment-516057341 -- it's offset subtracted from fd->offset to get offset into wbuffer.

app/Pasteboard.m
```diff
@@ -0,0 +1,247 @@
...
+        return fd_priv(fd).buffer;
+    }
+
+    if (check_read_generation(fd) != 0) {
+        return NULL;
+    }
+
+    NSString *contents = UIPasteboard.generalPasteboard.string;
+    *len = contents.length;
+    return contents.UTF8String;
```
tbodt: The UTF8String is freed when the NSString that it came from is freed, and that happens when this function returns. ASAN should warn you about this.

app/Pasteboard.m
```diff
@@ -0,0 +1,247 @@
...
+// buffer => UIPasteboard
+static int clipboard_wsync(clip_fd *fd) {
+    if (fd_priv(fd).buffer == NULL) {
+        return 0;
+    }
+
+    void *data = fd_priv(fd).buffer;
+    size_t len = fd_priv(fd).buffer_len;
+
+    // FIXME(stek29): This logs "Returning local object of class NSString"
```
tbodt: Do you still get this warning on the latest version of the PR?

app/Pasteboard.m
```diff
@@ -0,0 +1,247 @@
...
+        return _ENOMEM;
+    }
+
+    memcpy(fd_priv(fd).buffer, data, len);
+    fd_priv(fd).buffer_len = len;
+
+    return len;
+}
+
+static int clipboard_poll(clip_fd *fd) {
```
tbodt: This should just always return POLL_READ | POLL_WRITE, because reading and writing will never block. (The return from poll is defined as the operations that wouldn't block, not the operations that would succeed.)

app/Pasteboard.m
```diff
@@ -0,0 +1,247 @@
...
+    if (n != 0) {
+        memcpy(buf, data + fd->offset, n);
+        fd->offset += n;
+    }
+
+    return n;
+}
+
+static ssize_t clipboard_write(clip_fd *fd, const void *buf, size_t bufsize) {
+    size_t new_len = fd->offset + bufsize;
```
tbodt: This isn't quite right if you seek to the middle of the buffer and write something. Use the maximum of this and old_len.

app/Pasteboard.m
```diff
@@ -0,0 +1,247 @@
...
+    size_t old_len = fd_priv(fd).buffer_len;
+
+    if (realloc_to_fit(fd, new_len)) {
+        return _ENOMEM;
+    }
+
+    // fill the hole between new offset and old len
```
tbodt: Should fill in the hole between old_len and new_len, not old_len and fd->offset

app/Pasteboard.m
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    memcpy(fd_priv(fd).buffer + fd->offset, buf, bufsize);
+    fd->offset += bufsize;
+    fd_priv(fd).buffer_len = new_len;
+
+    return bufsize;
+}
+
+static off_t_ clipboard_lseek(clip_fd *fd, off_t_ off, int whence) {
```
tbodt: Not something you need to fix in this PR, but this logic is duplicated in procfs, so it should be moved to a utility function at some point.

fs/fd.h
```diff
@@ -55,6 +55,18 @@ struct fd {
...
             struct list unix_scm;
         } socket;
+
+        // See app/Pasteboard.m
+        struct {
+            // UIPasteboard.changeCount
+            uint64_t generation;
+            // Buffer for written data
+            void* buffer;
+            // it's capacity
```
tbodt: nit: "its" not "it's"

app/Pasteboard.m
stek29: I thought that variable was autoreleased for some reason.

app/Pasteboard.m
stek29: Agreed. I was thinking of making them blocking at first, but I haven't gotten to it.

app/Pasteboard.m
tbodt: Looks like the default is __strong, though you could use __autoreleasing to autorelease instead

app/Pasteboard.m
stek29: nope, it should feel between old_len and fd->offset since between offset and new_len is where buf gets copied to -- there's no point in memset'ing and then copying buffer over it.

tbodt: Ah, gotcha.

app/Pasteboard.m
stek29: Yup.

app/Pasteboard.m
stek29: Just checked -- unless I specifically wrap these 3 lines in `@autoreleasepool`, ASAN doesn't warn me about anything (because memory being referenced isn't actually freed).

stek29: Although I should probably add autoreleasepool's wherever I use `get_data` to free that data when it's no longer needed.

stek29: @tbodt sure, I'll add __autoreleasing, that makes sense.

stek29: judging by disasm default seems to be strong indeed.

tbodt:
Thanks!

PR #449, opened 2019-07-25
procfs: Fix short read offset handling (fix #448)

### LINKED ISSUE #448: procfs short read -- invalid offset
Current behavior:
```
vphone:~# exec 3<>/proc/version
vphone:~# head -c10 <&3 && echo
Linux vers
vphone:~# cat <&3 && echo
Linux version 3.2.0-ish SUPER AWESOME compiled on Ju
```

Expected behavior:
```
vphone:~# exec 3<>/proc/version
vphone:~# head -c10 <&3 && echo
Linux vers
vphone:~# cat <&3 && echo
ion 3.2.0-ish SUPER AWESOME compiled on Jul 25 2019
```

I've already fixed it, I'm just filling an issue to reference it.

stek29:


fs/proc.c
```diff
@@ -121,7 +121,7 @@ static ssize_t proc_read(struct fd *fd, void *buf, size_t bufsize) {
     if (n > remaining)
         n = remaining;
 
-    memcpy(buf, data, n);
+    memcpy(buf, data + fd->offset, n);
```
tbodt: oops

fs/proc.c
```diff
@@ -101,7 +101,7 @@ static int proc_refresh_data(struct fd *fd) {
     struct proc_entry entry = fd->proc.entry;
     ssize_t size = entry.meta->show(&entry, fd->proc.data);
     if (size < 0)
-        return size;
+        return (int)size;
```
tbodt: Why add this?

stek29: Xcode warns about implicit cast from ssize_t to smaller int (loses precision blah blah), I've added explicit cast to silence it, but didn't mean to add it to commit nor push it.

I've noticed when I've created PR, but it was not worth it to stash, switch back to proc-short-read branch, remove it, amend commit, force push and switch back to what I was working on.

Should I remove it?

tbodt: For some reason I haven't seen this warning before. I'd say it's fine then.

PR #453, opened 2019-07-29
Fixed scrolling

lorenzodla:
Scrolling fixed. Only one line of code but the "quid of the matter".

I was on the javascript and not on the iOS side.


tbodt:
Nice find! I'm pretty sure the touchmove listener is now not actually doing anything because you set it on the terminal instead of the terminal element. I tried just deleting it and things seemed to work. Could you test that on iOS 12 just to be sure?

lorenzodla:
Tested on iOS 12 and 13

tbodt:
I found a better way to fix scrolling, pushed to master. Thanks for this PR.

PR #475, opened 2019-09-05
Implement return cache optimization.

wishstudio:
This implements the return cache optimization for JIT mode.

The technical details of this optimization can be read at https://github.com/wishstudio/flinux/wiki/Dynamic-Binary-Translation#return-handling.

The implementation creates a new gadget named "call_postamble" and appends it after each call gadget. It checks whether the return address is correct and links to the actual successor code.

The actual performance improvement depends on application being run. I have observed run time reduction up to 15% on aarch64. No performance regression has been observed.

### DIFF
--- a/jit/gadgets-x86_64/control.S
+++ b/jit/gadgets-x86_64/control.S
@@ -8,40 +8,22 @@
     movl %r14d, (%_addrq)
     shrw $4, %r14w
     movzwl %r14w, %r14d
-    leaq 24(%_ip), %r15
-    movq %r15, LOCAL_ret_cache(%_cpu, %r14, 8)
+    movq %_ip, LOCAL_ret_cache(%_cpu, %r14, 8)
     write_done 32, call
-    movq 8(%_ip), %_ip
+    movq 32(%_ip), %_ip
     jmp jit_ret_chain
 
 .gadget call_indir
     subl $4, %_esp
     movl %_esp, %_addr
     write_prep 32, call_indir
-    movl 8(%_ip), %r14d
+    movl 16(%_ip), %r14d
     movl %r14d, (%_addrq)
     shrw $4, %r14w
     movzwl %r14w, %r14d
-    leaq 16(%_ip), %r15
-    movq %r15, LOCAL_ret_cache(%_cpu, %r14, 8)
+    movq %_ip, LOCAL_ret_cache(%_cpu, %r14, 8)
     write_done 32, call_indir
     movl %_tmp, %_eip
-    jmp jit_ret
-
-.gadget call_postamble
-    movl (%_ip), %r14d
-    movq 8(%_ip), %r15
-    cmpl %r14d, %tmpd
-    jnz 1f
-    movq 16(%_ip), %_ip
-    btq $63, %_ip
-    jc 1f
-    leaq -JIT_BLOCK_code(%_ip), %r15
-    movq %r15, LOCAL_last_block(%_cpu)
-    gret
-1:
-    movq %r15, LOCAL_last_block(%_cpu)
-    movl %tmpd, %_eip
     jmp jit_ret
 
 .gadget ret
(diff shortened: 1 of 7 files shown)


jit/gen.c
```diff
@@ -242,8 +242,8 @@ static inline bool gen_op(struct gen_state *state, gadget_t *gadgets, enum arg a
 #define jcc(cc, to, else) gagg(jmp, cond_##cc, to, else); jump_ips(-2, -1); end_block = true
 #define J_REL(cc, off)  jcc(cc, fake_ip + off, fake_ip)
 #define JN_REL(cc, off) jcc(cc, fake_ip, fake_ip + off)
-#define CALL(loc) load(loc, OP_SIZE); ggg(call_indir, saved_ip, fake_ip); gggg(call_postamble, fake_ip, state->block, fake_ip); jump_ips(-1, 0); end_block = true
-#define CALL_REL(off) gggg(call, saved_ip, fake_ip + off, fake_ip); gggg(call_postamble, fake_ip, state->block, fake_ip); jump_ips(-6, -1); end_block = true
+#define CALL(loc) load(loc, OP_SIZE); ggg(call_indir, saved_ip, fake_ip); gggg(call_postamble, fake_ip, -1, fake_ip); state->block->code[state->size - 2] = (long)state->block; jump_ips(-1, 0); end_block = true
+#define CALL_REL(off) gggg(call, saved_ip, fake_ip + off, fake_ip); gggg(call_postamble, fake_ip, -1, fake_ip); state->block->code[state->size - 2] = (long)state->block; jump_ips(-6, -1); end_block = true
```
tbodt: Why does this fix the bug?

jit/gadgets-x86_64/control.S
```diff
@@ -16,15 +20,44 @@
...
     read_prep 32, ret
-    movl (%_addrq), %_eip
+    movl (%_addrq), %tmpd
+    movl %tmpd, %r14d
+    shrw $4, %r14w
+    movzwq %r14w, %r14
+    movq LOCAL_ret_cache(%_cpu, %r14, 8), %_ip
+    cmpq $0, %_ip
+    jz 1f
+    gret
```
tbodt: Since this line will always go to a call_postamble gadget, it would make more sense to just write a jmp. It would then make more sense for call_postamble to be a function instead of a gadget, and make the call_postamble arguments into call arguments.

tbodt: Or make call_postamble a part of ret, and avoid the jump entirely.

In more detail: rearrange the arguments to call and call_indir so saved_ip, fake_ip, and current block are the first three. Current block is ignored by the call gadgets. ret updates _ip like it does now, then does the job of call_postamble, reading call arguments relative to _ip.

tbodt:
Thanks! I found a 10% improvement on vim startup time, which is not insignificant.

jit/gen.c
wishstudio: OK. Just found that this may not be the end of block realloc, because of the `gen_exit()` call later. We have to record this thing and fix at the end.

jit/gadgets-x86_64/control.S
wishstudio: OK. Makes sense to me.

wishstudio:
Addressed review comments.

jit/gen.c
```diff
@@ -47,6 +48,9 @@ void gen_end(struct gen_state *state) {
         list_init(&block->jumps_from[i]);
         list_init(&block->jumps_from_links[i]);
     }
+    if (state->block_patch_ip != 0) {
+        block->code[state->block_patch_ip] = (unsigned long)block;
```
tbodt: Add space between (unsigned long) and block please

jit/gen.c
```diff
@@ -118,6 +122,8 @@ typedef void (*gadget_t)(void);
 #define gg(_g, a) do { g(_g); GEN(a); } while (0)
 #define ggg(_g, a, b) do { g(_g); GEN(a); GEN(b); } while (0)
 #define gggg(_g, a, b, c) do { g(_g); GEN(a); GEN(b); GEN(c); } while (0)
+#define ggggg(_g, a, b, c, d) do { g(_g); GEN(a); GEN(b); GEN(c); GEN(d); } while (0)
+#define gggggg(_g, a, b, c, d, e) do { g(_g); GEN(a); GEN(b); GEN(c); GEN(d); GEN(e); } while (0)
```
tbodt: `gggggg` This is getting a little out of control. Just a tad. :D

(no need to change anything, just commenting)

wishstudio: We may consider change to variadic macros later. But to my experience some compilers don't like it.

tbodt:
Thanks!

Issue #435, opened 2019-07-22
iOS 13 multi window support


tbodt:
User feedback:
> It would be cool if you could have multiple instances of iSH open using iPad’s splitscreen/multitasking/slideover functionalities.

Mnpn03:
Has this been resolved by #545?

tbodt:
Yeah why isn't this closed

rluvaton:
Great work!
Is it possible to do it on iPhone?

kkk669:
@rluvaton No. It's exclusive for iPadOS.

PR #482, opened 2019-09-14
Multiwindow

NoahPeeters:
Adds support for multiple windows on iOS 13 while keeping support for single windows on iOS 12 and below. #435

For each new window a new tty is created in /dev/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,19 @@ macros:
       script:
           - meson build $MESON_OPTS
           - ninja -C build $NINJA_TARGET
+    - &mac
+      language: objective-c
+      install:
+          - pip3 install meson==0.49.2
+      addons:
+          homebrew:
+              packages:
+                  - ninja
+                  - llvm
+                  - yarn
+              update: true
+      script:
+          - xcodebuild -project iSH.xcodeproj -scheme iSH -sdk iphoneos CODE_SIGNING_ALLOWED=NO
 
 matrix:
     include:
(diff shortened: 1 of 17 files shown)


tbodt:
Wow thanks! Will look in more detail later, but one high level comment: please use pseudoterminals (ptys) so they appear in /dev/pts, which avoids having to constantly create and destroy terminal files. (This may require some hacks to the pseudoterminal code, which is OK with me.)

tbodt:
Also, the travis build has failed because it builds with Xcode 10. Is there some way to keep this building on Xcode 10 and only include the new features when built with Xcode 11?

NoahPeeters:
I just updated the code to only include the new feature if the iOS 13 sdk is present. That way the builds with Xcode 10 and the iOS 12 sdk still work and don't include the new features. I also added a new travis build with Xcode 11 in a separate commit. That works as well. If you don't want that, I can remove it. That does not seem to increase the build times on travis as the Xcode builds run in parallel and the new Xcode 11 build runs faster. It is probably a good idea to use pseudoterminals. I looked into that but I don't really know what changes are required for that. You already mount `devptsfs` in `/dev/pts` but I don't yet understand the file system completely.

tbodt:
Xcode stuff looks good. For pseudoterminals, looks like you'd find an empty slot in pty_slave.ttys, call tty_get on that slot, and replace the ops, I guess by separating ops into a separate field from driver. I'd add a new function to pty.c to do that, called something like pty_create_fake(const struct tty_ops *ops).

NoahPeeters:
Thanks a lot for the information. I haven't found the file pty.c because for some reason it is not included in the Xcode project. Now the whole file system logic makes more sense (there are some other files missing in the Xcode project as well). I will look into this.

tbodt:
The only reason there are any files outside of app included in the Xcode project is so I can set breakpoints on them. They aren't ever built by Xcode. I normally edit outside of Xcode.

dimthe:
Hello 

I am on iOS 12.4.1 on an iPad Pro . When the new iPados comes out in few days will i be able to use multiple tabs ?

theobouwman:
When will this be merged?

tbodt:
When it's done. There's a fair amount more work to do before it makes sense to merge this.

tbodt:
@NoahPeeters Are you planning to work on this more? If you'd like I can take it over.

NoahPeeters:
I didn't had much time to work on this. I was able to create the fake pseudoterminals but when I tried to start the start process and connect it to the tty the process quickly stopped. (Completely regular. It made a syscall to exit).

That's were I am currently stuck. If you like, you can take it over.

tbodt:
OK sounds good. Could you push your work on creating fake pseudoterminals somewhere? I'll take a crack next weekend.

tbodt:
I'll turn this into a new PR (but I'll make sure to credit you).

PR #517, opened 2019-10-13
Updated to Alpine 3.10.2

ELChris414:
There should be a way to download always the latest I believe.


ELChris414:
Guessing not too important at current time, closing PR.

PR #535, opened 2019-10-21
Update CPUID, Implement movq, pcmpeqb, orpd, pmovmskb

MatthewMerrill:
After these changes:
 - CPUID declares support for CMOV 
 - Implements movq, pcmpeqb, orpd, pmovskb
 - xmm/reg gadget has been added

SSE2 is NOT declared in CPUID, as that could trigger applications to opt into using yet unready sse instructions. I'd propose delaying enabling SSE2 at least until after node works with it.

PR #538, opened 2019-10-22
Add meson option for enabling SSE CPUID flag

MatthewMerrill:
Blocked on #535


MatthewMerrill:
Moved CPUID part of this to #746

PR #541, opened 2019-10-24
Download the latest alpine minirootfs

hrubymar10:
Don't hardcode latest alpine version, but download the latest one automatically instead.

### DIFF
--- a/iSH.xcodeproj/project.pbxproj
+++ b/iSH.xcodeproj/project.pbxproj
@@ -954,7 +954,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


filfat:
👏

tbodt:
This isn't a good idea because it gives the build a non-immutable dependency. Sometimes an Alpine upgrade breaks things, so I don't want it to happen implicitly. Similar to why people lock dependency versions with package-lock.json.

ntindle:
Should there be a way to run this explicitly without having to go through the terminal?

tbodt:
What do you mean?

ntindle:
More of a far in the future update to latest. I thought about it and it doesn’t really have a place right now to me

Issue #230, opened 2018-12-28
Mount folders from other file providers


jckarter:
It'd be awesome if you could open a folder in iCloud Drive, Working Copy, or other file providers and manipulate it directly from inside iSH as a mount point.

tbodt:
What would the UI look like for this? Probably it would be invoked through a share sheet, but then how do you specify where you want the file/folder mounted?

jckarter:
afaik it requires going through the open file UI to get access to a folder. In the (i) menu you could have a "Mount folder..." menu item that brings up the file picker. As far as where, maybe you could automatically map the folder under `/mnt/<folder name>`, or give the user another file picker to choose the mount location within the iSH filesystem.

emkayonline:
A good example of a UI for The is in the app LibTerm:

Using the folder icon one can use the iOS file browser to choose a folder (Using 'select', then choosing the item, then 'open')

This then becomes the current directory (which I use then for processing the files in that folder).

For ish, rather than becoming the current directory, it would make sense to bookmark the selected folder as a /mnt point. (The apps Scriptable and Pythonista bookmark opened folders so that then can be mounted across sessions)

tbodt:
This has been merged in now

PR #543, opened 2019-10-26
Add support for mounting folders from iOS

NoahPeeters:
Adds a button to the bar to mount any folder from a FileProvider into ish as described in #230.

Not all FileProvider support giving read/write access to a folder. Those aren't available.

Working FileProviders:
- On My iPad/iPhone
- iCloud Drive
- Working Copy
- SMB servers

FileProviders which do not work
- GoogleDrive
- ish itself
- OneDrive

### DIFF
--- a/app/SceneDelegate.h
+++ b/app/SceneDelegate.h
@@ -6,8 +6,11 @@
 //
 
 #import <UIKit/UIKit.h>
+#import "TerminalViewController.h"
 
 NS_ASSUME_NONNULL_BEGIN
+
+extern TerminalViewController *currentTerminalViewController;
 
 API_AVAILABLE(ios(13))
 @interface SceneDelegate : UIResponder <UIWindowSceneDelegate>
--- a/app/iOSFS.m
+++ b/app/iOSFS.m
@@ -1,0 +1,449 @@
+//
+//  iOSFS.m
+//  iSH
+//
+//  Created by Noah Peeters on 26.10.19.
+//
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+#include <sys/stat.h>
+#include "SceneDelegate.h"
+#include "iOSFS.h"
+#include "kernel/fs.h"
+#include "kernel/errno.h"
+#include "fs/path.h"
+
+const NSFileCoordinatorWritingOptions NSFileCoordinatorWritingForCreating = NSFileCoordinatorWritingForMerging;
+
+@interface DirectoryPickerDelegate : NSObject <UIDocumentPickerDelegate>
+
+@property NSURL *url;
+@property pthread_mutex_t *mutex;
+
+@end
+
+@implementation DirectoryPickerDelegate
+
+- (instancetype)initWithMutex:(pthread_mutex_t *)mutex {
+    if (self = [super init]) {
+        self.mutex = mutex;
+        self.url = nil;
+        pthread_mutex_lock(mutex);
+    }
+    return self;
+}
+
+- (void)documentPickerWasCancelled:(UIDocumentPickerViewController *)controller {
+    pthread_mutex_unlock(self.mutex);
+}
+
+- (void)documentPicker:(UIDocumentPickerViewController *)controller didPickDocumentsAtURLs:(NSArray<NSURL *> *)urls {
+    if (urls.count > 0)
+        self.url = urls[0];
+    pthread_mutex_unlock(self.mutex);
+}
+
+@end
+
+NSURL *url_for_mount(struct mount *mount) {
+    return (__bridge NSURL *) mount->data;
+}
+
+NSURL *url_for_path_in_mount(struct mount *mount, const char *path) {
+    return [[url_for_mount(mount) URLByAppendingPathComponent:[NSString stringWithUTF8String:path]] URLByStandardizingPath];
+}
+
+const char *path_for_url_in_mount(struct mount *mount, NSURL *url, const char *fallback) {
+    NSString *mount_path = [[url_for_mount(mount) URLByStandardizingPath] path];
+    NSString *url_path = [[url URLByStandardizingPath] path];
+
+    // The `/private` prefix is a special case as described in the documentation of `URLByStandardizingPath`.
+    if ([mount_path hasPrefix:@"/private/"]) mount_path = [mount_path substringFromIndex:8];
+    if ([url_path   hasPrefix:@"/private/"]) url_path   = [url_path   substringFromIndex:8];
+
+    if (![url_path hasPrefix:mount_path])
+        return fallback;
+
+    NSURL *new_url = [NSURL fileURLWithPath:[url_path substringFromIndex:[mount_path length]]];
+    return [new_url ? [new_url path] : @"" cStringUsingEncoding:NSUTF8StringEncoding];
+}
+
+static int iosfs_stat(struct mount *mount, const char *path, struct statbuf *fake_stat, bool follow_links);
+const struct fd_ops iosfs_fdops;
+
+static int posixErrorFromNSError(NSError *error) {
+    if (error != nil) {
+        NSError *underlyingError = [error.userInfo objectForKey:NSUnderlyingErrorKey];
+        if (underlyingError) {
+            return -(int)underlyingError.code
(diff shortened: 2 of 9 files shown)


tbodt:
Thanks! There are a few things missing, such as file coordination.

app/ExternalFolder.m
```diff
@@ -0,0 +1,56 @@
...
+
+    fs->name = "external";
+    fs->magic = 0x7265616d;
+    fs->mount = external_mount;
+    fs->umount = external_umount;
+
+    return fs;
+}
+
+const struct fs_ops *externalfs = nil;
```
tbodt: Can you create this statically to avoid having to call something to create it?

app/ExternalFolder.m
```diff
@@ -0,0 +1,56 @@
...
+
+NSURL* urlForMount(struct mount *mount) {
+    return (__bridge NSURL *) mount->data;
+}
+
+int external_mount(struct mount *mount) {
+    NSURL *url = urlForMount(mount);
+    if ([url startAccessingSecurityScopedResource] == NO) {
+        CFBridgingRelease(mount->data);
+        return errno_map();
```
tbodt: startAccessingSecurityScopedResource doesn't set errno so this will return garbage

fs/mount.c
```diff
@@ -37,15 +37,15 @@ void mount_release(struct mount *mount) {
     unlock(&mounts_lock);
 }
 
-int do_mount(const struct fs_ops *fs, const char *source, const char *point, int flags) {
+int do_mount_with_data(const struct fs_ops *fs, const char *source, const char *point, int flags, void *data) {
```
tbodt: You're supposed to set mount->data after calling this, there shouldn't need to be a new function

app/TerminalViewController.m
```diff
@@ -234,6 +238,48 @@ - (IBAction)pressArrow:(ArrowBarButton *)sender {
...
 
+- (IBAction)pressAddFolder:(id)sender {
+    UIDocumentPickerViewController *picker = [[UIDocumentPickerViewController alloc] initWithDocumentTypes:@[ @"public.folder"] inMode:UIDocumentPickerModeOpen];
+
+    picker.delegate = self;
+
+    [self presentViewController:picker animated:true completion:nil];
+}
+
+-(void)displayMountError {
```
tbodt: Should include some information on what the error is

tbodt:
Also I think a better idea for the UI is to open the UIDocumentPickerViewController code in the mount function. That way you type something like `mount -t ios website /mnt/website` and it pops up a file picker.

fs/mount.c
NoahPeeters: My problem was (and still is) that do_mount calls `fs->mount` and at that point `startAccessingSecurityScopedResource` must be called. For that, I need the NSURL object.

NoahPeeters: My problem was (and still is) that `do_mount` calls `fs->mount` and at that point `startAccessingSecurityScopedResource` must be called. For that, the `NSURL` object is required.

NoahPeeters:
I've updated the pull request. Now, new directories can be mounted with the `mount -t ...` command.
I also included the `NSFileCoordinator` but it is currently not used for opening files as I ran into some problems. One simple example is creating a fifo with mkfifo. If one now tries to read it with cat and then pipe data into the fifo with echo, the echo command blocks because it requires write access and cat still has read access to the file.

app/iOSFS.m
```diff
@@ -0,0 +1,320 @@
...
+
+    return err;
+}
+
+static struct fd *iosfs_open(struct mount *mount, const char *path, int flags, int mode) {
+    return realfs.open(mount, path, flags, mode);
+}
+
+static int iosfs_mount_data(char source[MAX_PATH], char point[MAX_PATH], dword_t *flags, void **data) {
+    pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```
tbodt: Can this be moved into iosfs_mount? I think that would eliminate the need for a new do_mount_with_data.

tbodt:
File coordination was clearly not designed for fifos. I think it's reasonable to check what type the file is, and only use file coordination for regular files where read/write can't block.

videoMonkey:
Would love to see this happen, is any progress being made? I wish my programming chops were at the level to be able to help.

NoahPeeters:
I had to do a lot for university lately. I will definitely continue working on this. Probably between Christmas and New Year or so. I already looked into it and it is not that hard to implement it.

videoMonkey:
Hey! Thanks!

—Jamie

NoahPeeters:
I continued working on this. I added file coordination for regular files and removed the `iosfs_mount_data` function by calling the ui functions in `iosfs_mount`.

I did a lot of testing with `git` in a mounted folder from Working Copy. I noticed that `git status` can be quit slow even with mid-size repositories because git will check every file in the repository and each of the coordinator calls slows down the process. That's why I added a second filesystem: `ios-unsafe`. That one will simply make sure that the correct permissions for the security scope are requested but the filecoordinators are not used at all. In case of Working Copy, this works very well but there are probably apps which can have a problem with that (that's why I called in `ios-UNSAFE`).

Therefore, this pull request now adds the following two commands:

```sh
mount -t ios website /mnt/website
mount -t ios-unsafe website /mnt/website
```

Finally, someone with more experience in async programming in objective-c should have a look at the usage of mutexes in `iosfs_open`. It does work but I do not know if there is a better/more elegant way of doing this.

tbodt:
Nice! Will do a detailed review in a bit.

I'm torn on whether ios-unsafe is a good idea. For the specific case of Working Copy, it makes more sense to just do git operations directly in the Working Copy app.

app/SceneDelegate.h
```diff
@@ -6,9 +6,12 @@
 //
 
 #import <UIKit/UIKit.h>
+#import "TerminalViewController.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
+extern TerminalViewController *currentTerminalViewController;
```
tbodt: I don't really like this way of getting the view controller. My other ideas are way more annoying to implement though.

app/iOSFS.m
```diff
@@ -0,0 +1,449 @@
...
+#include "kernel/fs.h"
+#include "kernel/errno.h"
+#include "fs/path.h"
+
+const NSFileCoordinatorWritingOptions NSFileCoordinatorWritingForCreating = NSFileCoordinatorWritingForMerging;
+
+@interface DirectoryPickerDelegate : NSObject <UIDocumentPickerDelegate>
+
+@property NSURL *url;
+@property pthread_mutex_t *mutex;
```
tbodt: Locks aren't really intended for waiting for something to happen. Should use a condition variable. cond_t from util/sync.h is preferred because it allows the process to be interrupted by a signal, with e.g. ctrl-c.

app/iOSFS.m
```diff
@@ -0,0 +1,449 @@
...
+    [coordinator coordinateWritingItemAtURL:src_url options:NSFileCoordinatorWritingForMoving error:&error byAccessor:^(NSURL *url) {
+        [coordinator itemAtURL:url willMoveToURL:dst_url];
+        err = realfs.rename(mount, path_for_url_in_mount(mount, url, src), dst);
+        [coordinator itemAtURL:url didMoveToURL:dst_url];
+    }];
+
+    return combine_error(error, err);
+}
+
+static int iosfs_symlink(struct mount *mount, const char *target, const char *link) {
```
tbodt: There's a lot of repeated code in these functions. How much can be factored out into a new function that takes a block?

tbodt:
@NoahPeeters Do you plan to work on this soon? If not, I'm willing to pick it up and get it merged.

tbodt:
@NoahPeeters Ping

ThisIsQasim:
@NoahPeeters really appreciate the work you have done so far and if you could get this to completion it would really supercharge my device and open it up for endless possibilities.

MrAlexWeber:
+1 for this feature, replying to bump and watch

tbodt:
Gonna take this over

tbodt:
Merged in dbd4100.

PR #545, opened 2019-10-28
Multiwindow

tbodt:
This is a pretty big change so I'd like some help testing it before I merge. If you're interested, please build and test this PR and make sure nothing weird or unexpected happens.

cc @NoahPeeters

### DIFF
--- a/app/AppDelegate.h
+++ b/app/AppDelegate.h
@@ -12,6 +12,8 @@
 @property (strong, nonatomic) UIWindow *window;
 - (void)exitApp;
 
++ (int)bootError;
+
 @end
 
 extern NSString *const ProcessExitedNotification;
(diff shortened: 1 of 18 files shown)


tbodt:
#482

NoahPeeters:
There seems to be an issue with network io. For example `ping 1.1.1.1` gets stuck after the first ping and one cannot stop it with CTRL-C. There is a similar behavior with `python -m SimpleHTTPServer`. It works until one tries to stop it. But this only happens for processes startet in the ptys. When switching to a default tty with cmd-option-shift-1 it works and it also works in tmux. And ping also seems to work after starting tmux, executing ping and exiting tmux again.

tbodt:
@NoahPeeters Fixed that. The signal mask was wrong.

PR #613, opened 2020-01-11
Fix pread/pwrite; make read/write be optional

saagarjha:
Written largely in GNU Nano and with minimal testing; in addition to the usual scrutiny I would check formatting if I were you.

### DIFF
--- a/kernel/fs.c
+++ b/kernel/fs.c
@@ -379,16 +379,18 @@ dword_t sys_pread(fd_t f, addr_t buf_addr, dword_t size, off_t_ off) {
             goto out;
         }
         res = fd->ops->read(fd, buf, size);
-        if (res >= 0) {
-            if (user_write(buf_addr, buf, res))
-                res = _EFAULT;
-        }
         // This really shouldn't fail. The lseek man page lists these reasons:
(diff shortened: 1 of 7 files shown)


kernel/fs.c
```diff
@@ -209,15 +209,25 @@ dword_t sys_read(fd_t fd_no, addr_t buf_addr, dword_t size) {
...
         res = _EBADF;
         goto out;
     }
     if (S_ISDIR(fd->type)) {
         res = _EISDIR;
         goto out;
     }
-    res = fd->ops->read(fd, buf, size);
+    if (fd->ops->pread) {
+        res = fd->ops->pread(fd, buf, size, fd->offset);
```
tbodt: This assigns an ssize_t to an int_t, I'd be worried about truncation

saagarjha: Both `pread(2)` and `read(2)` return a `ssize_t`, so I just kept the type that was there before:

https://github.com/tbodt/ish/blob/3ca3bf672a4cf2e0bc1a76e8eca83aef47d3d79b/kernel/fs.c#L220

What should I be using? `int_t`?

tbodt: Fair. Let's leave this then.

PR #619, opened 2020-01-18
Upgrade Alpine to 3.11.3

0b101:
https://alpinelinux.org/posts/Alpine-3.11.0-released.html & https://alpinelinux.org/posts/Alpine-3.11.3-released.html

Due to a problem with Xcode and my 10 year old Macbook, I was unable to test this. I did, however, update Alpine manually on my iPhone and I didn't experience any errors.


tbodt:
Don't want to bother for a point release, as an apk add will get the new versions of packages regardless.

Issue #47, opened 2018-11-11
GDB does not work


ghanan94:
I tried to install gdb with the hope of debugging why tmux wouldn’t work (another issue I submitted) and found out that gdb doesn’t work either :(

\# apk add gdb
...
\# gdb tmux
Segmentation Fault
\# gdb -h
Segmentation Fault
\# gdb
Segmentation Fault
\#

tbodt:
I've never tried to run gdb so I'm not surprised this happens. In any case, gdb relies on ptrace which is totally unimplemented.

Mnpn03:
```
Device:~# apk add gdb
OK: 92 MiB in 45 packages
Device:~# gdb
Segmentation fault
Device:~# dmesg
FIXME socketcall 16
13 page fault on 0x50 at 0x567336fd
```

ntindle:
`34 missing syscall 136` When debugging hello world. 
```c
#include <stdio.h>
int main()
{
   printf("Hello, World!");
   return 0;
}
```

If saved as `prog1.c`

```bash
gcc  -Wall -Werror -ansi -pedantic-error
s -g prog1.c -o prog1.x
gdb prog1.x
```

```
Reading symbols from prog1.x...
(gdb) run

saagarjha:
Should be fixed by #736, so closing this for now. Much of ptrace is still unimplemented, but you should be able to debug simple single-threaded programs now.

PR #642, opened 2020-02-17
Add syscall 136 personality

krystophny:
This provides the simple part of fixing #47, as GDB makes system call 136 `sys_personality` with arguments `0x0000` (set and return value for Linux) and `0xffffffffff` (get value). In any other case error code -1 is returned. See http://man7.org/linux/man-pages/man2/personality.2.html

### DIFF
--- a/kernel/getset.c
+++ b/kernel/getset.c
@@ -170,3 +170,12 @@ int_t sys_capset(addr_t header_addr, addr_t data_addr) {
     STRACE("capset(%#x, %#x)", header_addr, data_addr);
     return 0;
 }
+
+// minimal version according to Linux sys/personality.h
+dword_t sys_personality(dword_t pers) {
+    if (pers == 0xffffffff)  // get personality, return default (Linux)
+        return 0x00000000;
+    if (pers == 0x00000000)  // set personality to Linux
+        return 0x00000000;
+    return -1;  // otherwise return error
+}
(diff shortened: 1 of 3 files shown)


kernel/getset.c
```diff
@@ -170,3 +170,12 @@ int_t sys_capset(addr_t header_addr, addr_t data_addr) {
...
     return 0;
 }
+
+// minimal version according to Linux sys/personality.h
+dword_t sys_personality(dword_t pers) {
+    if (pers == 0xffffffff)  // get personality, return default (Linux)
+        return 0x00000000;
+    if (pers == 0x00000000)  // set personality to Linux
+        return 0x00000000;
+    return -1;  // otherwise return error
```
tbodt: The kernel doesn't just return -1 on error, it returns a negative version of the errno. You can find these in kernel/errno.h.

krystophny: Allright - error should be _EINVAL and I've also corrected the return type to a signed int. BTW, on usual Linux (Arch Linux, x86_64), just any input works and sets personality to some value, so if something still fails with the syscall one could extend it later. Example:

```C
#include <sys/personality.h>
#include <stdio.h>

void main() {
  printf("testing personalities...\n");
  for (unsigned long i=0x00; i<=0xff; i++) {
    int ret = personality(i);
    printf("previous: %d, current: %d\n", ret, i);
  }
  printf("current: %d\n", personality(0xffffffff));
}
```

krystophny:
Great, thanks for merging. iSH is one of the most amazing pieces of software I've ever seen, so keep up the great work!

Issue #628, opened 2020-01-31
kerberos commands return “Bad file descriptor”


lyon-fnal:
HI - I’m really looking forward to use ish to access computing resources from my iPad. We use kerberos for security. I did 

```
apk add krb5
```

and copied my `krb5.conf` file to `/etc/krb5.conf`.

I get a Kerberos ticket with `kinit`. The communication with our kerberos server seems to work, but it ends with an error...

```bash
ipad-132427:~# kinit <removed>
Password for <removed> 
kinit: Failed to store credentials: Bad file descriptor (filename: /tmp/krb5cc_0) while getting initial credentials
```

There’s an option to `kinit` that saves credentials to any file, and when I do a file in `/root` it fails the same way.

Furthermore, the `klist` command, which tells me what tickets I have (maybe none) also fails with a similar Bad file descriptor error. 

I imagine the kerberos executables were built in some incompatible way. Any ideas? Let me know how I can help!  Thanks, Adam

thunderkeys:
I can replicate, but I wanted to add that the credential cache IS generated properly after authenticating - if it is copied to another (regular, non-iSH) system, it works with klist/can be used to authenticate to other systems.

```
examplesystem:~ exampleuser$ KRB5CCNAME=FILE:/Users/exampleuser/ccache klist
Credentials cache: FILE:/Users/exampleuser/ccache
        Principal: exampleuser@EXAMPLE.COM

  Issued                Expires               Principal
Feb 25 09:10:56 2020  Feb 25 19:10:56 2020  krbtgt/EXAMPLE.COM@EXAMPLE.COM

[ssh somewhere]

examplesystem:~ exampleuser$ KRB5CCNAME=FILE:/Users/exampleuser/ccache klist
Credentials cache: FILE:/Users/exampleuser/ccache
        Principal: exampleuser@EXAMPLE.COM

  Issued                Expires               Principal
Feb 25 09:15:17 2020  Feb 25 19:10:56 2020  host/somesystem.second.example.com@SECOND.EXAMPLE.COM
Feb 25 09:15:17 2020  Feb 25 19:10:56 2020  krbtgt/SECOND.EXAMPLE.COM@EXAMPLE.COM
Feb 25 09:10:56 2020  Feb 25 19:10:56 2020  krbtgt/EXAMPLE.COM@EXAMPLE.COM
```

tbodt:
[...]

PR #649, opened 2020-02-26
Fix lockfile flag checks

thunderkeys:
Lock flags fix for #628 (kerberos commands return "Bad file descriptor")


tbodt:
Did a little cleanup on the conditional, would probably be good to quickly test whether it still works before I merge.

thunderkeys:
Still working with those updates.

tbodt:
Thanks again!

PR #664, opened 2020-03-23
translate README.md to  a chinese Version

ChaosTong:
add README_ZH.md, a translate of README.md to  a chinese version


tbodt:
Thanks! I'm worried about this becoming out of date though, since the english version is the main one. I guess I can add a link in the readme with a disclaimer.

PR #683, opened 2020-04-16
Xcode Build Phases

RandyMcMillan:
Change: Create Alpine Filesystem

Add: Modify /etc/motd phase

  1. We isolate this action for clarity

Add: Modify /etc/profile

  1. In this phase we have added an example on how to install the git package
     if not already installed.
     It is commented out to not add over head to CI builds

  2. We remove the /etc/motd file after the initial setup
     This is a way to not redundantly see the motd at every boot.
     We also use the hash cat output to test if /etc/motd not exist
     This provides a block for persistant configs only  after
     the initial boot and setup.

### DIFF
--- a/iSH.xcodeproj/project.pbxproj
+++ b/iSH.xcodeproj/project.pbxproj
@@ -792,6 +792,8 @@
 				BB792B4D1F96D90D00FFB7A4 /* Frameworks */,
 				BBF1248B1FA7BF530088FB50 /* Download Alpine */,
 				BBF1248A1FA7BDBA0088FB50 /* Create Alpine Filesystem */,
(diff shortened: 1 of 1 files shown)


RandyMcMillan:
Screen Shot 2020-04-16 at 6 56 56 PM

Screen Shot 2020-04-16 at 6 57 05 PM

Screen Shot 2020-04-16 at 6 57 46 PM
Screen Shot 2020-04-16 at 6 57 56 PM

tbodt:
I don't want to make significant modifications to the Alpine base system. Even if that were a good idea, the place to do it would not be Xcode build phases.

PR #712, opened 2020-05-10
CPU refactor, sine cera

saagarjha:


Rexogamer:
so the idea is to split things into separate prs? fair

saagarjha:
Actually the idea is to finish the implementation this time

util/sync.h
```diff
@@ -5,6 +5,7 @@
 #include <pthread.h>
 #include <stdbool.h>
 #include <setjmp.h>
+#include <sys/errno.h>
```
tbodt, line 8: Shouldn't this just be errno.h?

saagarjha: Probably

PR #724, opened 2020-05-17
Terminate mount after dismissing a document picker

### LINKED ISSUE #703: Terminate mount command automatically after cancelling the document picker
When I tap or click the outside of the document picker, it will be closed with the `mount` command still running. So, I have to press Ctrl+C to terminate it.

Please see the following video.

https://youtu.be/MTSSxKa6qgk

Is it possible to terminate the `mount` command automatically after cancelling the document picker?

kkk669:
Fixes #703

### DIFF
--- a/app/iOSFS.m
+++ b/app/iOSFS.m
@@ -17,7 +17,7 @@
 
 const NSFileCoordinatorWritingOptions NSFileCoordinatorWritingForCreating = NSFileCoordinatorWritingForMerging;
 
-@interface DirectoryPicker : NSObject <UIDocumentPickerDelegate>
+@interface DirectoryPicker : NSObject <UIDocumentPickerDelegate, UIAdaptivePresentationControllerDelegate>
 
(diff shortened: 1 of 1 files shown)


tbodt:
Thanks for finding this!

Could you make sure it works on iOS 12? The API you're using says it was added in iOS 13.

kkk669:
@tbodt I completely forgot to test in iOS 12. I will check that tomorrow.

kkk669:
@tbodt I tested on the iOS 12.4 simulator and the 11.4 one and both of them worked well. `UIAdaptivePresentationControllerDelegate` itself has been added since iOS 8, so my modifications can be compiled for iOS 12. Besides, iOS 12.4 or earlier cannot dismiss a modal sheet like iOS 13, so it doesn't need `presentationControllerDidDismiss`.

tbodt:
Looks good. Thanks!

PR #732, opened 2020-05-23
Make iOSFS work on iOS 11 and 12 (#700)

### LINKED ISSUE #700: Mounting not working on iOS 12
iPad Air, running iOS 12.4.6. Looks like the file picker is coming up instead of the directory picker. 

iPad picker

Working as expected on an iPhone 6s, iOS 13.3.1.

kkk669:
I added `picker.allowsMultipleSelection = YES` when it's running on iOS 12 or earlier. This enables the iOSFS to work on iOS 12 or earlier. Besides, behaviors on iOS 13 will not be changed.

If multiple directories are selected, the first one is mounted as you can see on the video below.

IMB_8zX63v

Fixes #700


tbodt:
Thanks!

Issue #532, opened 2019-10-20
cursor doesn’t change color when on characters


wenmin-wu:
cursor doesn't change color or be transparent when on characters. This makes the character under the cursor can't be seen at all. The same when I use vim, even I configured it in vimrc.
![DDBFD2FD-9F5D-4548-823C-4535212C724C](https://user-images.githubusercontent.com/9409333/67158259-77965500-f368-11e9-836c-208e33229b1e.jpeg)

tbodt:
Duplicate of https://crbug.com/1010197

tbodt:
Should leave this open for potential contributors. With one contribution, you can help iSH and Chrome OS at the same time!

wenmin-wu:
> Should leave this open for potential contributors. With one contribution, you can help iSH and Chrome OS at the same time!

what's the progress of #57 ?
Compared to this issue, I more like to contribute to the `MMX` problem. Since it blocks me from using `kubectl` on my iPad.

PR #733, opened 2020-05-23
Make cursor blink, mitigates #532

oflebbe:
Found no easy way to reverse color the cursor, but blinking makes the text readable. At least with a 50% chance :)


tbodt:
A switch in settings to make the cursor blink would be a good idea, but I don't like having it blink all the time. I just added code to hterm to put the character under the cursor into the cursor node, which looks fine.

shzadakam (approved):

PR #736, opened 2020-05-27
ptrace

saagarjha:
It's back


app/AppGroup.h
```diff
@@ -5,4 +5,4 @@
 //  Created by Theodore Dubois on 2/28/20.
 //
 
-NSURL *ContainerURL();
+NSURL *ContainerURL(void);
```
tbodt, line 8: Test for whether you can review a draft PR

saagarjha:
What kinda works:

* Launching GDB
* Starting a process
* Setting symbolic breakpoints in libc
* Setting breakpoints at an address
* Reading registers
* Single stepping

What does not work:

* You tell me, but I will probably be unsurprised

fs/proc/pid.c
```diff
@@ -111,7 +132,7 @@ static int proc_pid_cmdline_show(struct proc_entry *entry, struct proc_data *buf
         goto out_free_task;
 
     size_t size = task->mm->argv_end - task->mm->argv_start;
-    uint8_t *data = malloc(size);
+    char *data = malloc(size);
```
tbodt, line 135: What's wrong with uint8_t?

kernel/calls.c
```diff
@@ -272,6 +275,20 @@ void handle_interrupt(int interrupt) {
...
         deliver_signal(current, SIGILL_, info);
+    } else if (interrupt == INT_BREAKPOINT) {
+        lock(&pids_lock);
+        send_signal(current, SIGTRAP_, (struct siginfo_) {
+            .sig = SIGTRAP_,
+            .code = SI_KERNEL_,
+        });
+        unlock(&pids_lock);
+    } else if (interrupt == INT_DEBUG) {
+        lock(&pids_lock);
```
tbodt, line 286: Shouldn't need to lock pids_lock, as the signal is being sent to current?

kernel/calls.c
```diff
@@ -272,6 +275,20 @@ void handle_interrupt(int interrupt) {
...
+        send_signal(current, SIGTRAP_, (struct siginfo_) {
+            .sig = SIGTRAP_,
```
tbodt, line 288: Shouldn't be needed
https://github.com/ish-app/ish/blob/95dc3f5290efcd90fe009dab63cd19bcdfe32ca4/kernel/signal.c#L60

kernel/exec.c
```diff
@@ -616,6 +616,17 @@ int __do_execve(const char *file, struct exec_args argv, struct exec_args envp)
 
     current->did_exec = true;
     vfork_notify(current);
+
+    if (current->traced) {
+        lock(&pids_lock);
```
tbodt, line 621: unnecessary?

kernel/task.h
```diff
@@ -78,6 +88,8 @@ struct task {
     cond_t *waiting_cond;
     lock_t *waiting_lock;
     lock_t waiting_cond_lock;
+
+    bool traced;
```
tbodt, line 92: Would this make more sense inside of .ptrace?

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
...
+                unlock(&child->ptrace.stop_lock);
+            }
+        }
+    }
+    return child;
+}
+
+dword_t sys_ptrace(dword_t request, dword_t pid, addr_t addr, dword_t data) {
+    switch (request) {
+    case PTRACE_TRACEME_:
```
tbodt, line 26: Please indent the case label and case body by one more level, and add a blank line after each case, to be consistent with the rest of the project

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
...
+        STRACE("ptrace(PTRACE_TRACEME, %d, %#x, %#x)", pid, addr, data);
+        current->traced = true;
+        return 0;
+    case PTRACE_PEEKTEXT_:
+    case PTRACE_PEEKDATA_: {
+        STRACE("ptrace(PTRACE_PEEKDATA, %d, %#x, %#x)", pid, addr, data);
+        dword_t peek;
+        struct task *child = find_child(current, pid);
+        if (child) {
```
tbodt, line 35: Should be `if (child == NULL) return _EPERM;` to avoid having to indent the rest of the case (same with other cases)

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
...
+            if (user_get_task(child, addr, peek)) {
+                unlock(&child->ptrace.stop_lock);
+                return _EFAULT;
+            } else if (user_put(data, peek)) {
```
tbodt, line 39: Dunno if `else` is needed here, given the return statement in the preceding if

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
...
+                // child->cpu.xds = user_regs_.xds;
+                // child->cpu.xes = user_regs_.xes;
+                // child->cpu.xfs = user_regs_.xfs;
+                // child->cpu.xgs = user_regs_.xgs;
+                // child->cpu.eax = user_regs_.orig_eax;
+                child->cpu.eip = user_regs_.eip;
+                // child->cpu.xcs = user_regs_.xcs;
+                child->cpu.eflags = user_regs_.eflags;
+                child->cpu.esp = user_regs_.esp;
+                // child->cpu.xss = user_regs_.xss;
```
tbodt, line 185: This is duplicated in POKEUSER, pull into a function? Same with PEEKUSER and GETREGS

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
...
+        return 0;
+    }
+    case PTRACE_SETFPREGS_: {
+        STRACE("ptrace(PTRACE_SETFPREGS, %d, %#x, %#x)", pid, addr, data);
+        struct task *child = find_child(current, pid);
+        if (child) {
+            struct user_fpregs_struct_ user_regs_;
+            if (user_get(data, user_regs_)) {
+                return _EFAULT;
+            } else {
```
tbodt, line 215: oops

kernel/ptrace.h
```diff
@@ -0,0 +1,56 @@
...
+    dword_t xgs;
+    dword_t orig_eax;
+    dword_t eip;
+    dword_t xcs;
+    dword_t eflags;
+    dword_t esp;
+    dword_t xss;
+};
+
+typedef struct user_fpregs_struct_ {
```
tbodt, line 43: Why a typedef? It seems unused

kernel/exit.c
```diff
@@ -281,6 +289,11 @@ int do_wait(int idtype, pid_t_ id, struct siginfo_ *info, struct rusage_ *rusage
         info->child.pid = id;
         if (reap_if_needed(task, info, rusage, options))
             goto found_zombie;
+        if (task->ptrace.stopped && task->ptrace.signal) {
+            info->child.status = task->ptrace.trap_event << 16 | task->ptrace.signal << 8 | 0x7f;
+            task->ptrace.signal = 0;
+            goto found_ptrace_stopped;
+        }
```
tbodt, line 296: Could this be moved into reap_if_needed to avoid duplication?

kernel/exit.c
```diff
@@ -243,7 +246,7 @@ static bool reap_if_needed(struct task *task, struct siginfo_ *info_out, struct
 int do_wait(int idtype, pid_t_ id, struct siginfo_ *info, struct rusage_ *rusage, int options) {
     if (idtype != P_ALL_ && idtype != P_PID_ && idtype != P_PGID_)
         return _EINVAL;
-    if (options & ~(WNOHANG_|WUNTRACED_|WEXITED_|WCONTINUED_|WNOWAIT_))
+    if (options & ~(WNOHANG_|WUNTRACED_|WEXITED_|WCONTINUED_|WNOWAIT_|__WALL_))
```
tbodt, line 249: Should there be something checking for and handling `__WALL_`, or is it OK to ignore?

emu/memory.h
```diff
@@ -113,6 +114,7 @@ int pt_copy_on_write(struct mem *src, struct mem *dst, page_t start, page_t page
 #define MEM_WRITE 1
 // Must call with mem read-locked.
 void *mem_ptr(struct mem *mem, addr_t addr, int type);
```
tbodt, line 116: How about a new value for `type` (maybe `MEM_WRITE_PTRACE`) instead of a new boolean argument?

kernel/fork.c
```diff
@@ -172,6 +173,13 @@ dword_t sys_clone(dword_t flags, addr_t stack, addr_t ptid, addr_t tls, addr_t c
 
     // task might be destroyed by the time we finish, so save the pid
     pid_t pid = task->pid;
+
+    if (current->traced) {
+        current->ptrace.trap_event = PTRACE_EVENT_FORK_;
+        lock(&pids_lock);
```
tbodt, line 179: Missing unlock(&pids_lock), but I dunno if the lock is necessary (see other comments)

kernel/task.h
```diff
@@ -47,6 +47,16 @@ struct task {
     sigset_t_ saved_mask;
     bool has_saved_mask;
 
+    struct {
+        bool stopped;
+        bool reported_stop;
```
tbodt, line 52: Appears to be unused

kernel/task.h
```diff
@@ -47,6 +47,16 @@ struct task {
...
+        cond_t stop_cond;
+        lock_t stop_lock;
```
tbodt, line 54: I think this should be named just `lock`, so `ptrace.lock` protects all the things that can be modified through ptrace. Also add a comment saying what it locks

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
+#include "ptrace.h"
+#include "kernel/calls.h"
+#include "kernel/errno.h"
+#include "kernel/signal.h"
+#include "task.h"
+#include <string.h>
+
+// Returns stopped child with the given pid, locked with stop_lock
+static struct task *find_child(struct task *parent, pid_t_ pid) {
```
tbodt, line 9: Is it necessary to have the task parameter if it's always `current`?

kernel/signal.c
```diff
@@ -97,13 +97,38 @@ void deliver_signal(struct task *task, int sig, struct siginfo_ info) {
...
+
 void send_signal(struct task *task, int sig, struct siginfo_ info) {
     // signal zero is for testing whether a process exists
     if (sig == 0)
         return;
     if (task->zombie)
         return;
 
+    if (task->traced && sig != SIGKILL_) {
+        signal_delivery_stop(task, sig, info);
```
tbodt, line 128: Shouldn't this be called from the receiving side of the signal, not the sending side? I think if you use `kill` on a ptraced process this would try to put the `kill` in signal-delivery stop.

fs/proc/pid.c
saagarjha: You know, if `CHAR_BIT != 8`

kernel/calls.c
saagarjha: `signal_delivery_stop` requires `pids_lock`: https://github.com/ish-app/ish/blob/f1eb124ee7559c83bdc584ac9c671b9203c39a27/kernel/signal.c#L109

fs/proc/pid.c
tbodt: Is that really allowed?

C99 section 5.2.4.2.1:
> Their implementation-defined values shall be equal or greater in magnitude (absolute value) to those shown, with the same sign.
> `CHAR_BIT` = 8: number of bits for smallest object that is not a bit-field (byte)

section 7.18.1.1:
> The typedef name `uint`*N*`_t` designates an unsigned integer type with width *N*, no padding bits, and a two’s complement representation. Thus, `uint24_t` denotes a signed integer type with a width of exactly 24 bits.
> These types are optional.

So `CHAR_BIT` must be at least 8, and can only be greater than 8 if no 8-bit non-bitfield type exists. `uint8_t` is optional, but if it does exist, then `CHAR_BIT` is required to be 8.

kernel/calls.c
saagarjha: I didn't have that at first, but IIRC this was necessary because the signal is swallowed by `signal_delivery_stop` and made available via `ptrace(PTRACE_GETSIGINFO, …)`, at which point the signal needed to be correct.

kernel/exec.c
saagarjha: Taking the lock, or sending a signal? I think both are necessary…

kernel/exit.c
saagarjha: I think in recent kernel versions this is implied or something like that, but GDB passes it in anyways to be safe so we just silently accept it as a valid flag.

kernel/exec.c
tbodt: Taking the lock, but no need to discuss it here, there are like 500 other comments about it

kernel/exit.c
saagarjha: I feel like there was a reason I put it here but I might have refactored it so that it doesn't matter anymore. I'll give it a try.

kernel/ptrace.c
saagarjha: Well, there's multiple problems with it. One is that yet, the parameter is useless (I'll remove it), and another is that you can trace things that aren't your children. But since we don't have attach yet this function "works" for now.

kernel/ptrace.c
saagarjha: Just to confirm: one is `addr`, one is `data`.

kernel/ptrace.c
saagarjha: probably a good idea

kernel/ptrace.c
saagarjha: yeah I should probably put a comment there, GDB needs this to return success before it lets you execute anything

kernel/ptrace.h
saagarjha: it is, will remove

kernel/signal.c
saagarjha: Uh, yes, I think it should. This might break a whole lot of stuff though so I'll have to look into it.

kernel/task.h
saagarjha: Will remove if they are

kernel/task.h
saagarjha: Yeah, I think it should be protected by the lock too.

kernel/exit.c
saagarjha: Err, no, I figured out why. I need to go to `found_ptrace_stopped`. I guess I could put the condition and that code into `reap_if_needed` and return false and then check again but that seemed a bit ugly…

kernel/signal.c
tbodt: Hopefully it also fixes a lot of stuff

fs/proc/pid.c
saagarjha: Yeah but something something aliasing

kernel/exit.c
tbodt: Could you put the `info->sig = SIGCHLD_` in some kind of if statement so it doesn't happen if it's not supposed to? Or put it in reap_if_zombie?

kernel/ptrace.c
```diff
@@ -0,0 +1,243 @@
...
+        if (child->pid == pid) {
+            lock(&child->ptrace.lock);
+            if (child->ptrace.stopped) {
+                break;
+            } else {
+                unlock(&child->ptrace.lock);
+            }
+        }
+    }
+    return child;
```
tbodt, line 21: Just noticed what I think is a bug here: if none of the children have the given PID, it will return the last child, not NULL.

abdullah0sedeq:
و

kernel/group.c
```diff
@@ -59,12 +59,15 @@ dword_t sys_setpgrp() {
 
 pid_t_ sys_getpgid(pid_t_ pid) {
     STRACE("getpgid(%d)", pid);
-    if (pid != 0 && pid != current->pid)
-        return _EPERM;
     lock(&pids_lock);
-    pid_t_ pgid = current->group->pgid;
+    struct task *task;
+    if (!(task = !pid ? current : pid_get_task(pid))) {
```
tbodt, line 64: too dense! aaaaaa

emu/memory.c
```diff
@@ -256,10 +256,13 @@ void *mem_ptr(struct mem *mem, addr_t addr, int type) {
...
     }
 
-    if (entry != NULL && type == MEM_WRITE) {
+    if (entry != NULL && (type == MEM_WRITE || type == MEM_WRITE_PTRACE)) {
         // if page is unwritable, well tough luck
-        if (!(entry->flags & P_WRITE))
+        if (type != MEM_WRITE_PTRACE && !(entry->flags & P_WRITE))
             return NULL;
+        if (type == MEM_WRITE_PTRACE) {
+            entry->flags |= P_WRITE | P_COW;
```
tbodt, line 264:
```suggestion
            // TODO: Is P_WRITE really correct? The page shouldn't be writable without ptrace.
            entry->flags |= P_WRITE | P_COW;
```

kernel/group.c
saagarjha: no u

tbodt: OK, I can fix after merging

kernel/ptrace.c
```diff
@@ -0,0 +1,260 @@
...
+// Ensure stopped, ptrace locked, etc. before calling this
+static void get_user_regs(struct cpu_state *cpu, struct user_regs_struct_ *user_regs_) {
+   user_regs_->ebx = cpu->ebx;
+   user_regs_->ecx = cpu->ecx;
+   user_regs_->edx = cpu->edx;
+   user_regs_->esi = cpu->esi;
+   user_regs_->edi = cpu->edi;
+   user_regs_->ebp = cpu->ebp;
+   user_regs_->eax = cpu->eax;
+// user_regs_->xds = cpu->xds;
```
tbodt, line 35: Indentation on these comments is a bit funky compared to set_user_regs

emu/memory.c
saagarjha: I'm unsure what the concern is here?

tbodt: I think if P_WRITE is set that will transmute a read-only page into a page that can be written by userspace, which doesn't seem right. Doesn't need to be fixed right away, hence the suggestion of a TODO.

PR #737, opened 2020-05-27
Fix things

saagarjha:
Not ptrace-specific

### DIFF
--- a/app/AppGroup.m
+++ b/app/AppGroup.m
@@ -41,14 +41,14 @@ static NSDictionary *AppEntitlements() {
         return nil;
     if (dl_info.dli_fbase == NULL)
         return nil;
-    uint8_t *base = dl_info.dli_fbase;
+    char *base = dl_info.dli_fbase;
     struct mach_header_64 *header = dl_info.dli_fbase;
     if (header->magic != MH_MAGIC_64)
         return nil;
     
     // Simulator executables have fake entitlements in the code signature. The real entitlements can be found in an __entitlements section.
     size_t entitlements_size;
-    uint8_t *entitlements_data = getsectiondata(header, "__TEXT", "__entitlements", &entitlements_size);
+    char *entitlements_data = getsectiondata(header, "__TEXT", "__entitlements", &entitlements_size);
     if (entitlements_data != NULL) {
         NSData *data = [NSData dataWithBytesNoCopy:entitlements_data
                                             length:entitlements_size
@@ -67,7 +67,7 @@ static NSDictionary *AppEntitlements() {
             cs_lc = (void *) lc;
             break;
         }
-        lc = (void *) ((uint8_t *) lc + lc->cmdsize);
+        lc = (void *) ((char *) lc + lc->cmdsize);
     }
     if (cs_lc == NULL)
         return nil;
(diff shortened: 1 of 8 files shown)


kernel/task.c
```diff
@@ -94,6 +94,11 @@ void (*task_run_hook)(void) = NULL;
 
 static void *task_run(void *task) {
     current = task;
+#if __APPLE__
```
tbodt, line 97: Could you move the ifdefs into a function, maybe in sync.h, so they don't have to be duplicated for both fork and exec?

util/sync.h
```diff
@@ -40,12 +40,13 @@ static inline void __lock(lock_t *lock, __attribute__((unused)) const char *file
...
 static inline void unlock(lock_t *lock) {
-    pthread_mutex_unlock(&lock->m);
 #if LOCK_DEBUG
     lock->debug = (struct lock_debug) {};
 #endif
+    lock->owner = NULL;
+    pthread_mutex_unlock(&lock->m);
 }
-#define unlock(lock) pthread_mutex_unlock(&(lock)->m)
+#define unlock(lock) unlock(lock)
```
tbodt, line 49: Don't think there's a reason to have this as a macro anymore

kernel/exit.c
```diff
@@ -220,8 +220,10 @@ static bool reap_if_zombie(struct task *task, struct siginfo_ *info_out, struct
 }
 
 static bool notify_if_stopped(struct task *task, struct siginfo_ *info_out) {
-    // FIXME the check of task->group->stopped needs locking I think
-    if (!task->group->stopped || task->group->group_exit_code == 0)
+    lock(&current->group->lock);
```
tbodt, line 223: Should be task->group->lock

app/AppGroup.m
```diff
@@ -101,11 +101,11 @@
                                                                       error:nil];
 }
 
-NSArray<NSString *> *CurrentAppGroups() {
+NSArray<NSString *> *CurrentAppGroups(void) {
     return AppEntitlements()[@"com.apple.security.application-groups"];
 }
 
-NSURL *ContainerURL() {
+NSURL *ContainerURL(void) {
```
tbodt, line 108: Pretty sure (void) isn't needed for function definitions, only prototypes

saagarjha: It is not, although trying to call it incorrectly invokes undefined behavior

tbodt: The idea is you need it on the prototype so you get an error when you call it wrong, but you don't need it on the definition

saagarjha: Apparently defining it as such creates a 0-argument prototype as well, and then trying to call a function that has a prototype defined with the wrong number of arguments is undefined.

tbodt: Weird but ok

app/AppGroup.m
```diff
@@ -41,14 +41,14 @@
         return nil;
     if (dl_info.dli_fbase == NULL)
         return nil;
-    uint8_t *base = dl_info.dli_fbase;
+    char *base = dl_info.dli_fbase;
```
tbodt, line 44: What does this fix?

saagarjha: `char *` and `void *` are the legal aliasing types

tbodt: Guilty as charged

saagarjha: I should note that *any* character type is a legal aliasing type, so you are also welcome to use `signed char *` and `unsigned char *` if you so wish.

PR #772, opened 2020-06-09
Fix locking issues with CPU refactor

tbodt:
There was a deadlock when cpu_run called cpu_step_to_interrupt with mem
read-locked, and it would write-lock mem to clean up jetsam.


tbodt:
Tests pass, how bad can it be

PR #790, opened 2020-06-20
Turn on Ligatures

CRKatri:
Very self explanatory


tbodt:
Thanks!

Reference for posterity: https://chromium.googlesource.com/apps/libapps/+/master/nassh/doc/FAQ.md#are-font-ligatures-supported

PR #800, opened 2020-06-25
Remove the headers build phase from libarchive

iamrecursion:
1.  The headers build phase was causing the 'archive' build to produce
    an Xcode generic archive rather than an application archive during
    packaging. This PR removes that build phase from the libarchive
    subproject.


tbodt:
Thanks for the PR!

PR #804, opened 2020-06-27
Updated e2e.bash

### LINKED ISSUE #744: e2e tests produce div by zero error message when test isn't found
```
ish (master) $ bash tests/e2e/e2e.bash -e foobar
Runtime error (func=(main), adr=15): Divide by zero
### Passed: 0/0 (%)                ###
```

OrangeFran:
The script now does not fail if no tests were run. The crash occurred because the script tried to divide by $NUM_TOTAL even though this would be 0.

fixes #744


tbodt:
Thanks for fixing this!

Issue #59, opened 2018-11-12
Feature requeat: Siri Shortcuts integration


balboah:
I know the focus right now is compatibility with the basic userland, but what about making it possible to pipe inputs and outputs through Siri shortcuts a.k.a. Workflows?
It would make it possible to ”Shell out” from any app supporting it without the need of a remote server. 

First Idea that comes to mind:
- Open example code website
- share action -> Shortcuts app
- choose a workflow with action “run iSH command”
- exec command with clipboard pasted as stdin 
- return stdout back to the shortcuts workflow

Raupinger:
I think it would be much more efficient to do x-callback url support first. You could still integrate iSH in Workflows, but would be much more flexible.

tbodt:
bump

Tomoli75:
bump

bderiso:
+1 (sorry, i’m on mobile so i’m not sure how to best “vote” for issues here).

Using command line on a mobile device on-the-go requires extra focus on scripting and automation, so be able to launch common scripts using Shortcuts is ideal.

creepinson:
I could probably try and work on this as well, but I was already working on getting audio done.

NorthIsUp:
This would be so amazing +1

icearith:
I wonder is this feature under development?

TheGamer1002:
Bump! 🙂

saagarjha:
As far as I am aware, there is no work underway to support this feature at the moment. It's still open if anyone wants to jump in and implement it, but efforts from the us (that is, the core development team) are currently focused elsewhere.

ifuchs:
Has anyone ever found/created a way to invoke a shortcut from ish?

ninjamar:
bump

PR #842, opened 2020-07-26
WIP: Add support for using iSH on Apple Watches

AlexGustafsson:
# Intent

The intent of this PR is to start development for supporting iSH on Apple Watches.

iSH is a fantastic piece of code which I think is quite unprecedented. By implementing support for watchOS, the project could continue to grow and allow for new use cases.

# Goals
Goals:
* Support basic iSH functionality on WatchOS such as running commands and seeing their output

Non-goals:
* Fully interactive shell due to quite obvious interaction constraints on watches

# Planned approach

The watch app will be developed using Swift and SwiftUI. The app is meant to run side by side with the iSH app and act as a simple UI for the app.

Within the main iOS app a view will be added to support saving commands such as `nslookup google.com` for easy access within the watchOS app. The watchOS app will then allow for running these commands and see their output.

The app is planned to interact with the same layer as the terminal running within the iOS app to enable shared interactions.

# Comments
I hope to continue to work on this PR every now and then, but without any real time constraints.

I welcome any comments on code quality, procedures etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ You'll need these things to build the project:
 
 Open the project in Xcode, open iSH.xcconfig, and change `ROOT_BUNDLE_IDENTIFIER` to something unique. Then click Run. There are scripts that should do everything else automatically. If you run into any problems, open an issue and I'll try to help.
 
+If you want to deploy the project onto your own devices, you'll need a valid developer account. Sign in to your account via Xcode: Xcode -> Preferences -> Accounts. Now go to the project settings and change the value for Signing > Development Team under Build Settings to your development team. To do this a bit more automatically, change the value for one project and then change all occurances of `CK5SXRTBR7` in `iSH.xcodeproj/project.pbxproj` to your development ID for all `DevelopmentTeam` fields. Your ID will be the one not matching `CK5SXRTBR7`.
+
 ## Build command line tool for testing
 
 To set up your environment, cd to the project and run `meson build` to create a build directory in `build`. Then cd to the build directory and run `ninja`.
--- a/app/Info.plist
+++ b/app/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
 	<key>CFBundleVersion</key>
-	<string>73</string>
+	<string>$(PRODUCT_VERSION)</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
--- a/iSH.xcconfig
+++ b/iSH.xcconfig
@@ -5,3 +5,6 @@ ROOT_BUNDLE_IDENTIFIER = app.ish.iSH
 ISH_LOG =
 
 PRODUCT_APP_GROUP_IDENTIFIER = group.$(ROOT_BUNDLE_IDENTIFIER)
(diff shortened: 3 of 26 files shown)


tbodt:
I'm not entirely convinced this is a good idea, but I guess we'll find out.

Also it's best if you use objective-c instead of swift, the app is small enough that adding an empty swift file would inflate it by 4x.

filfat:
I think that this could be extremely interesting as a way to run certain scripts on the go. E.g. maybe call an API, restart or issue commands to a game server etc etc.

As an example imagine running a server for a game at home and you'd like to see current online players. You could have a ./query_players.sh script that rcons the server and prints the players on the watch. Or maybe it's crashed and you have a script that connects to the server through SSH and starts it again.

I would actually find this really useful personally @tbodt

AlexGustafsson:
@tbodt I’m not sure if how representative it is, but for me the iSH app went from 4.9MB to 8.5MB according to the systems settings. It is quite bad in relative terms, but not quite 5X. I think the situation has gotten a lot better recently since Apple is now bundling most things related to Swift within iOS, which they didn’t do for quite some time (if I remember things correctly).

I’ll try to get the MVP running with Swift and SwiftUI due to its simplicity and then port it to ObjC once we know what we want to achieve.

As for the use cases you’re describing, @filfat, that’s more or less exactly what I’m trying to solve. It mostly works now, but there’s no way of simply add commands to show up on the watch.

The extension of this is to get the Shortcuts app to work with iSH, but that’s an issue for another day.

tbodt:
Also iSH supports iOS 11, so SwiftUI is out. But it's ok to start with SwiftUI and then port it to ObjC/UIKit.

AlexGustafsson:
Thanks for the explanation, @tbodt, Swift is definitely out then.

dontbanmeplz:
What about preset commands that can be changed from the app

AlexGustafsson:
@dontbanmeplz I was thinking about that as well and that's what I'm currently doing in the WIP. However, with the arrival of the Shortcut app to the watch, perhaps it's better to investigate if one can handle commands from Shortcuts instead. Does anyone know if it's possible to add custom abilities so that iSH can be invoked from Shortcuts? If so, we might kill two birds with one stone by implementing that support instead.

dontbanmeplz:
@AlexGustafsson if you could make it so you could execute commands and see their output from shortcuts app the shortcuts community would now have no bounds and they would thank you. if it is possible PLEASE do that.

tbodt:
See #59. I looked into this once but at the time the shortcut would have to execute in a different process which would make it pretty unworkable. iOS 14 removes this limitation with in-app intent handling, but I haven't had the chance to give it another shot. Would appreciate contributions here.

dontbanmeplz:
Tbh i know nothing about iOS coding so i cant be of much help

ifuchs:
You can do this now but it is a bit ugly. Start sshd, put it in the
background, use the Run Script over ssh Shortcut action, et voilà. It even
works in personal automations with no user intervention required.

dontbanmeplz:
Well yes but you know it would be so much better

0n1cOn3:
Hey ! 

Interesting approach. But There was no update even after 3 years. So I would like to ask what the current stage this WIP is and if this will be continued or dropped.

Thanks in advance
0n1c0n3

PR #970, opened 2020-10-25
Fix build on Apple Silicon hosts

### LINKED ISSUE #913: Command ExternalBuildToolExecution failed with a nonzero exit code (library not found)
I tried building iSH with Xcode 12(12A7209) and macOS Big Sur (20A5374g), but Xcode build failed with error Command ExternalBuildToolExecution failed with a nonzero exit code. The full log indicated that a library was not found.

Sanity check compile stdout:
-----
Sanity check compile stderr:
ld: library not found for -lSystem
clang: error: linker command failed with exit code 1 (use -v to see invocation)
-----
meson.build:1:0: ERROR: Compiler env -u SDKROOT -u IPHONEOS_DEPLOYMENT_TARGET clang can not compile programs.

I'm using meson 0.53.0. Does anyone have a clue on how to solve this?

tbodt:
Invoking clang directly through the Xcode toolchain (which happens because
Xcode sets the PATH) instead of through /usr/bin/clang seems to break SDK
autodetection. Use xcrun to get around this.

Fixes #913

### DIFF
--- a/xcode-meson.sh
+++ b/xcode-meson.sh
@@ -4,7 +4,7 @@ cd $MESON_BUILD_DIR
 
 config=$(meson introspect --buildoptions)
 if [[ $? -ne 0 ]]; then
-    export CC_FOR_BUILD="env -u SDKROOT -u IPHONEOS_DEPLOYMENT_TARGET clang"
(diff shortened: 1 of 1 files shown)


saagarjha (approved):
LGTM.

PR #1004, opened 2020-10-30
Clean up some things

saagarjha:


saagarjha:
Where's the CI?

app/AppDelegate.m
```diff
@@ -225,7 +225,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
         // For iOS <13, where the app delegate owns the window instead of the scene
         if ([NSUserDefaults.standardUserDefaults boolForKey:@"recovery"]) {
             UINavigationController *vc = [[UIStoryboard storyboardWithName:@"About" bundle:nil] instantiateInitialViewController];
-            AboutViewController *avc = vc.topViewController;
+            AboutViewController *avc = (AboutViewController *)vc.topViewController;
```
tbodt, line 228: In this codebase we put spaces after casts
```suggestion
            AboutViewController *avc = (AboutViewController *) vc.topViewController;
```

emu/float80-test.c
```diff
@@ -5,6 +5,9 @@
 #include <fenv.h>
 #include "float80.h"
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wliteral-range"
```
tbodt, line 9: Is this a clang-only warning? If not, use GCC pragmas. If so, I don't see any need to push when you just pop at the end of the file

fs/tmp.c
```diff
@@ -436,20 +436,22 @@ static off_t_ tmpfs_lseek(struct fd *fd, off_t_ off, int whence) {
     }
 
     // TODO: dedupe code with procfs
```
tbodt, line 438: Please fix

fs/generic.c
```diff
@@ -287,3 +287,22 @@ int generic_rmdirat(struct fd *at, const char *path_raw) {
...
+int generic_seek(struct fd *fd, off_t_ off, int whence) {
+    off_t_ new_off = fd->offset;
+    if (whence == LSEEK_SET) {
+        fd->offset = off;
+    } else if (whence == LSEEK_CUR) {
+        if (new_off += off < 0)
+            return _EINVAL;
+        fd->offset = new_off;
+    } else if (whence == LSEEK_END) {
+        new_off = fd->proc.data.size + off;
```
tbodt, line 300: size needs to be an argument I think, this will break tmpfs as is

fs/generic.c
```diff
@@ -287,3 +287,22 @@ int generic_rmdirat(struct fd *at, const char *path_raw) {
     mount_release(mount);
     return err;
 }
+
+int generic_seek(struct fd *fd, off_t_ off, int whence) {
+    off_t_ new_off = fd->offset;
+    if (whence == LSEEK_SET) {
+        fd->offset = off;
+    } else if (whence == LSEEK_CUR) {
+        if (new_off += off < 0)
```
tbodt, line 296: I've been putting off fixing this because I think it's a little bit harder to fix than this. Something like "off < 0 && fd->offset > old_off"

app/AppGroup.m
```diff
@@ -35,20 +39,11 @@
...
-    char *base = dl_info.dli_fbase;
-    struct mach_header_64 *header = dl_info.dli_fbase;
-    if (header->magic != MH_MAGIC_64)
-        return nil;
+    const struct mach_header_64 *header = &_mh_execute_header;
     
     // Simulator executables have fake entitlements in the code signature. The real entitlements can be found in an __entitlements section.
     size_t entitlements_size;
-    char *entitlements_data = getsectiondata(header, "__TEXT", "__entitlements", &entitlements_size);
+    char *entitlements_data = (char *)getsectiondata(header, "__TEXT", "__entitlements", &entitlements_size);
```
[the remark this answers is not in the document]
saagarjha: What did I miss

PR #1038, opened 2020-11-11
Fix KVO crashes

tbodt:


app/NSObject+SaneKVO.m
```diff
@@ -0,0 +1,72 @@
...
+    KVOBlock newBlock = ^{
+        id owner = weakOwner;
+        if (owner) {
+            block(owner);
+        }
+    };
+    @synchronized (owner) {
+        for (NSString *keyPath in keyPaths) {
+            NSMutableSet *observations = objc_getAssociatedObject(owner, kKVOObservations);
+            if (observations == nil) {
```
saagarjha, line 36: `!observations`

app/NSObject+SaneKVO.h
```diff
@@ -0,0 +1,35 @@
...
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^KVOBlock)(void);
+
+@interface KVOObservation : NSObject {
+    BOOL _enabled;
+    __weak id _object;
```
saagarjha, line 16: Possibly `id<NSKeyValueObserving>`?

app/NSObject+SaneKVO.h
```diff
@@ -0,0 +1,35 @@
...
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^KVOBlock)(void);
+
+@interface KVOObservation : NSObject {
+    BOOL _enabled;
```
saagarjha, line 15: Is this used?

app/NSObject+SaneKVO.m
```diff
@@ -0,0 +1,72 @@
...
+    return observation;
+}
+
+- (void)observe:(NSArray<NSString *> *)keyPaths options:(NSKeyValueObservingOptions)options owner:(id)owner usingBlock:(void (^)(id self))block {
+    __weak id weakOwner = owner;
+    KVOBlock newBlock = ^{
+        id owner = weakOwner;
+        if (owner) {
+            block(owner);
+        }
```
saagarjha, line 31: This sounds like a good time to disable early

app/Roots.m
```diff
@@ -53,18 +53,21 @@ - (instancetype)init {
         [self addObserver:self forKeyPath:@"roots" options:0 context:nil];
```
saagarjha, line 53: Can we fix this?

saagarjha (commented):
Initial comments

app/NSObject+SaneKVO.h
saagarjha: Nevermind

app/NSObject+SaneKVO.h
```diff
@@ -0,0 +1,35 @@
...
+
+@interface NSObject (SaneKVO)
+
+- (KVOObservation *)observe:(NSString *)keyPath
+                    options:(NSKeyValueObservingOptions)options
+                 usingBlock:(KVOBlock)block;
+- (void)observe:(NSArray<NSString *> *)keyPaths
+        options:(NSKeyValueObservingOptions)options
+          owner:(id)owner
+     usingBlock:(void (^)(id self))block;
```
saagarjha, line 31: These should be prefixed

app/NSObject+SaneKVO.m
saagarjha: Actually, I would assert here, this should never happen…

app/NSObject+SaneKVO.h
tbodt: Yes, in -disable

app/NSObject+SaneKVO.h
tbodt: but I don't wanna

app/NSObject+SaneKVO.m
tbodt: I like this better, also it doesn't really matter

app/Roots.m
tbodt: Sure

app/NSObject+SaneKVO.m
tbodt: Done

app/NSObject+SaneKVO.h
saagarjha: But you should.

app/NSObject+SaneKVO.h
saagarjha: My actual concern is that this never seems to be set to `YES`

tbodt: There needs to be some way to mark this as disabled so it won't be disabled a second time. Also maybe someday I'll add -enable.

app/NSObject+SaneKVO.h
```diff
@@ -0,0 +1,35 @@
...
+//  iSH
+//
+//  Created by Theodore Dubois on 11/10/20.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^KVOBlock)(void);
```
saagarjha, line 12: Does this really need a `typedef`

app/NSObject+SaneKVO.h
saagarjha: My concern is that `-disable` seems to be a no-op at the moment

tbodt: Oops

tbodt: Oops

app/NSObject+SaneKVO.h
tbodt: Not really

app/NSObject+SaneKVO.h
```diff
@@ -0,0 +1,33 @@
...
+
+@interface NSObject (SaneKVO)
+
+- (KVOObservation *)observe:(NSString *)keyPath
+                    options:(NSKeyValueObservingOptions)options
+                 usingBlock:(void (^)(void))block;
+- (void)observe:(NSArray<NSString *> *)keyPaths
+        options:(NSKeyValueObservingOptions)options
+          owner:(id)owner
+     usingBlock:(void (^)(id self))block;
```
saagarjha, line 29: `owner`

app/Roots.m
```diff
@@ -50,21 +51,28 @@ - (instancetype)init {
...
+}
+- (void)setDefaultRoot:(NSString *)defaultRoot {
+    [NSUserDefaults.standardUserDefaults setObject:defaultRoot forKey:kDefaultRoot];
+}
+
 - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary<NSKeyValueChangeKey,id> *)change context:(void *)context {
-    if ([keyPath isEqualToString:@"defaultRoot"]) {
-        [NSUserDefaults.standardUserDefaults setObject:self.defaultRoot forKey:kDefaultRoot];
-    } else if ([keyPath isEqualToString:@"roots"]) {
+    if ([keyPath isEqualToString:@"roots"]) {
```
saagarjha, line 75: Can we delete this code?

saagarjha (commented):
Last couple things I noticed.

app/Roots.m
tbodt: Yes

app/NSObject+SaneKVO.h
tbodt: The argument is self because it is typical for the caller to want to name it self

saagarjha: You're welcome to call it `self` if that is the thing you are setting as the owner, but the header should at least match the semantics of the code (which is generic).

tbodt: Changed to leave out the argument name

Issue #273, opened 2019-01-12
htop does not run


weiszr:
I successfully install htop, but nothing is shown (no screen output at all).

Mrc527:
I can confirm, same behavior on iOS 13 as well.

weiszr:
Yup...still does not work.



______________________________________________________________
Robert Weiss
*Professor of Natural Hazards*
Department of Geosciences
Virginia Tech
4044 Derring Hall (0420)
Blacksburg, VA 24061, U.S.A.

Office: 1068A Derring
Phone:+1-540-231-2334
Fax: +1-540-231-3386


Research Web: http://coastalhazardsvt.weebly.com
Education Web: https://fralin.vt.edu/DRRM.html


Twitter: @VTCoastal

--


--

LinuxOnboard:
I believe it is a problem with ncurses.

LinuxOnboard:
What are some ncurses commands/programs that you tested?

jaderfeijo:
I don’t believe this is related to ncurses, I’ve been able to install and use tmux which so far has worked for me.

Assfugil:
it won't respond to SIGKILL either.

StewAlexander-com:
Anyone really good with stack traces? Maybe this might shed light on what's going on with htop?

62f:
@tbodt, hello. I made these changes:

- Loading per cpu usage
- List of alive pids
- More `TASK_MAY_BLOCK`

__Originally posted by @nimelehin in https://github.com/ish-app/ish/pull/1066#issuecomment-742613261__

https://discord.com/channels/508839261924229141/510973900209913857/811252086557114378 htop also gives me ```No btime in /proc/stat: No such file or directory``` for Alpine 3.13.1

yes htop and non-busybox tops are broken

62f:
![BFE151AC-70F3-4E3A-9E82-B698289F679F](https://user-images.githubusercontent.com/68332197/117942991-aae95d80-b2d9-11eb-96d0-f9dc408c7384.jpeg)

duckfromdiscord:
im having this issue too, build 1.2.3 298

PR #1066, opened 2020-11-16
WIP: Procfs enhancements

nimelehin:


nimelehin:
Hello @tbodt! Asking you to review these changes, they make `htop` working and also extend procfs. 

The patches can collect statistics on which tasks are blocked. This is done using a special wrapper in zones which can block our code (e.g calling real's os syscall). The wrapper is implemented with `for` which is strange, but I've checked that compiler can easily unroll this loop, so no performance degradation and it looks nice in C, I think.

Also updated procfs, so `htop` no longer freezing.
Screenshot 2020-11-18 at 17 07 02

nimelehin:
Seems it closes #273

tbodt:
Thanks for the PR! The one thing I'm not sure about is the blocking regions, as that's a pretty large code change and it's easy to miss a spot. I guess nothing really bad happens if you do, though.

kernel/task.h
```diff
@@ -188,4 +194,26 @@ extern void (*exit_hook)(struct task *task, int code);
...
 void update_thread_name(void);
 
+// To collect statics on which tasks are blocked we need to proccess areas
+// of code which could block our task (e.g reads or writes). Before executing
+// of functions which can block the task, we mark our task as blocked and
+// unblock it after the function is executed.
+__attribute__((always_inline)) inline int enter_blocking_area(void) {
+    lock(&pids_lock);
+    current->io_block = 1;
+    blocked_task_count++;
```
tbodt, line 204: Instead of maintaining a counter which could get out of sync easily, I'd prefer to iterate over the process list and count the number of running processes when you read the proc file. There aren't going to be all that many of them. This goes for both `blocked_task_count` and `alive_task_count`.

kernel/task.h
```diff
@@ -188,4 +194,26 @@ extern void (*exit_hook)(struct task *task, int code);
...
+
+__attribute__((always_inline)) inline int leave_blocking_area(void) {
+    lock(&pids_lock);
+    current->io_block = 0;
+    blocked_task_count--;
+    unlock(&pids_lock);
+    return 0;
+}
+
+#define TASK_BLOCKING_AREA for(int i = enter_blocking_area(); i < 1; leave_blocking_area(), i++)
```
tbodt, line 217: What do you think of naming this `TASK_MAY_BLOCK`? Then `enter_blocking_area`/`leave_blocking_area` becomes `task_may_block_start()` and `task_may_block_end()`

tbodt, line 217:
```suggestion
#define TASK_BLOCKING_AREA for (int i = enter_blocking_area(); i < 1; leave_blocking_area(), i++)
```

kernel/signal.c
```diff
@@ -636,8 +636,10 @@ int_t sys_rt_sigsuspend(addr_t mask_addr, uint_t size) {
 
 int_t sys_pause() {
     lock(&current->sighand->lock);
-    while (wait_for(&current->pause, &current->sighand->lock, NULL) != _EINTR)
-        continue;
+    TASK_BLOCKING_AREA {
```
tbodt, line 639: Instead of doing this, it would make more sense to put the blocking area inside of wait_for.

kernel/fs.c
```diff
@@ -461,6 +473,7 @@ dword_t sys_pwrite(fd_t f, addr_t buf_addr, dword_t size, off_t_ off) {
         return _ENOMEM;
     if (user_read(buf_addr, buf, size))
         return _EFAULT;
+    enter_blocking_area();
```
tbodt, line 476: I think this can use the macro? There aren't any gotos in the way like sys_pread has.

fs/proc/root.c
```diff
@@ -13,8 +15,25 @@ static int proc_show_version(struct proc_entry *UNUSED(entry), struct proc_data
 }
 
 static int proc_show_stat(struct proc_entry *UNUSED(entry), struct proc_data *buf) {
+    int ncpus = get_cpu_count();
     struct cpu_usage usage = get_cpu_usage();
-    proc_printf(buf, "cpu  %"PRIu64" %"PRIu64" %"PRIu64" %"PRIu64"\n", usage.user_ticks, usage.nice_ticks, usage.system_ticks, usage.idle_ticks);
+    struct uptime_info uptime_info = get_uptime();
+    unsigned uptime = uptime_info.uptime_ticks;
+    
+    uint64_t one_cpu_user_ticks = usage.user_ticks / ncpus;
```
tbodt, line 23: Is there an API that allows collecting these stats for the individual CPUs?

tbodt (commented):
A slight problem here is some of these stats are for the whole system including any iOS daemons that are running and some of these stats are only for processes running inside iSH. This could get a bit confusing in htop if you can see CPU usage for everything but only see some processes in the breakdown. It's probably fine though.

kernel/signal.c
nimelehin: Unfortunately this might bring a deadlock. I just want this blocking defines as simple as possible, so I just put them on the most of blocking syscalls in Linux ([list of blocking sys calls](https://techpubs.jurassic.nl/manuals/0650/developer/Debugger_Ref/sgi_html/ch10.html)). From this list only fcntl and close are not wrapped with `TASK_MAY_BLOCK`, since their implementation in our case is not blocking.

tbodt: This list seems outdated? It doesn't include futex, for example.

emkey1:
As an FYI, this patch fixes the procps top command as well.

IMG_0110

nimelehin:
@tbodt, hello. I made these changes:

- Loading per cpu usage
- List of alive pids
- More `TASK_MAY_BLOCK`

rbreaves:
Can we get this merged? htop is broken till it is..

emkey1:
The reliance on locking/unlocking  pids_lock in the TASK_MAY_BLOCK macro is a bit problematic from a performance/reliability perspective as so many other parts of iSH use it as well.

me355:
Shah

PR #1081, opened 2020-11-22
Add apk server using ODRs

tbodt:
TODO:
- [ ] UI for slow downloads
- [ ] ~~/etc/ish-version~~
- [x] Disable server when build flag is unset
- [x] Figure out why vim doesn't work
- [ ] ~~Consider installing custom signing keys~~
- [x] Don't put APKINDEX in the same pack as other stuff
- [ ] ~~More intelligent batching (e.g. dependency aware)~~
- [x] Stream the resource from disk through HTTP instead of loading it into memory
- [x] Remove tqdm dependency

### DIFF
--- a/app/APKServer.h
+++ b/app/APKServer.h
@@ -1,0 +1,16 @@
+//
+//  APKServer.h
+//  iSH
+//
+//  Created by Theodore Dubois on 11/21/20.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface APKServer : NSObject
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/app/APKServer.m
+++ b/app/APKServer.m
@@ -1,0 +1,139 @@
+//
+//  APKServer.m
+//  iSH
+//
+//  Created by Theodore Dubois on 11/21/20.
+//
+
+#import <CFNetwork/CFNetwork.h>
+#import <sys/socket.h>
+#import <netinet/in.h>
+#import "APKServer.h"
+
+@interface APKServer ()
+
+@property NSFileHandle *server;
+@property CFMutableDictionaryRef messages;
+
+@end
+
+@implementation APKServer
+
+- (instancetype)init {
+    if (self = [super init]) {
+        self.messages = CFDictionaryCreateMutable(kCFAllocatorDefault, 0, &kCFTypeDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks);
+        [self start];
+    }
+    return self;
+}
+
+- (void)start {
+    CFSocketRef socket = CFSocketCreate(kCFAllocatorDefault, AF_INET, SOCK_STREAM, 0, 0, NULL, NULL);
+    if (socket == nil) {
+        NSLog(@"APK server: failed to create socket");
+        return;
+    }
+    int one = 1;
+    if (setsockopt(CFSocketGetNative(socket), SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one)) != 0) {
+        NSLog(@"APK server: failed to set SO_REUSEADDR");
+        return;
+    }
+    struct sockaddr_in sin = {
+        .sin_len = sizeof(sin),
+        .sin_family = AF_INET,
+        .sin_port = htons(42069), // highly nice
+        .sin_addr = INADDR_ANY,
+    };
+    NSData *sin_data = [NSData dataWithBytes:&sin length:sizeof(sin)];
+    if (CFSocketSetAddress(socket, (CFDataRef) sin_data) != kCFSocketSuccess) {
+        NSLog(@"APK server: failed to bind");
+        return;
+    }
+    self.server = [[NSFileHandle alloc] initWithFileDescriptor:CFSocketGetNative(socket) closeOnDealloc:YES];
+    [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(onAccept:) name:NSFileHandleConnectionAcceptedNotification object:self.server];
+    [self.server acceptConnectionInBackgroundAndNotify];
+    NSLog(@"APK server: serving");
+}
+
+- (void)onAccept:(NSNotification *)n {
+    NSFileHandle *client = n.userInfo[NSFileHandleNotificationFileHandleItem];
+    CFDictionaryAddValue(self.messages, (__bridge CFTypeRef) client, CFHTTPMessageCreateEmpty(kCFAllocatorDefault, YES));
+    [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(onReadable:) name:NSFileHandleDataAvailableNotification object:client];
+    [client waitForDataInBackgroundAndNotify];
+    [self.server acceptConnectionInBackgroundAndNotify];
+}
+
+- (void)onReadable:(NSNotification *)n {
+    NSFileHandle *client = n.object;
+    CFHTTPMessageRef message = (CFHTTPMessageRef) CFDictionaryGetValue(self.messages, (__bridge CFTypeRef) client);
+    NSData *data = [client availableData];
+    if (data.length == 0)
+        return [self endRequest:client];
+    if (!CFHTTPMessageAppendBytes(message, data.bytes, data.length))
+        return [self endRequest:client];
+    if (CFHTTPMessageIsHeaderComplete(message))
+        return [self sendResponseForRequest:message toClient:client];
+    [client waitForDataInBackgroundAndNotify];
+}
+
+- (void)sendResponseForRequest:(CFHTTPMessageRef)message toClient:(NSFileHandle *)client {
+    NSURL *url = CFBridgingRelease(CFHTTPMessageCopyRequestURL(message));
+    NSArray<NSString *> *components = url.pathComponents;
+    if ([components.firstObject isEqualToString:@"/"])
+        components = [components subarrayWithRange:NSMakeRange(1, components.count - 1)];
+    NSString *resourceTag = [components componentsJoinedByString:@":"];
+
+    NSLog(@"APK server: fetching %@", resourceTag);
+    NSBundleResourceRequest *request = [[NSBundleResourceRequest alloc] initWithTags:[NSSet setWithObject:resourceTag]];
+    request.loadingPriority = NSBundleResourceRequestLoadingPriorityUrgent;
+    [request beginAccessingResourcesWithCompletionHandler:^(NSError * _Nullable error) {
+        int status = 200;
+        NSData *body = nil;
+        NSFileHandle *bodyFile = nil;
+        NSUInteger contentLength = 0;
+
+        if (error == nil) {
+            NSLog(@"APK server: serving %@", resourceTag);
+            bodyFile = [NSFileHandle fileHandleForReadingFromURL:[request.bundle URLForResource:resourceTag withExtension:nil] error:&error];
+            [bodyFile seekToEndOfFile];
+            contentLength = bodyFile.offsetInFile;
+            [bodyFile seekToFileOffset:0];
+        }
+        if (error != nil) {
+            NSLog(@"APK server: failed to fetch %@: %@", resourceTag, error);
+            status = 500;
+            body = [error.description dataUsingEncoding:NSUTF8StringEncoding];
+        }
+
+        CFHTTPMessageRef response = CFHTTPMessageCreateResponse(kCFAllocatorDefault, status, NULL, kCFHTTPVersion1_1);
+        if (body) {
+            CFHTTPMessageSetBody(response, (__bridge CFDataRef) body);
+            contentLength = body.length;
+        }
+        CFHTTPMessageSetHeaderFieldValue(response, CFSTR("Content-Length"), (__bridge CFStringRef) [NSString stringWithFormat:@"%lu", contentLength]);
+        [client writeData:CFBridgingRelease(CFHTTPMessageCopySerializedMessage(response))];
+        CFRelease(response);
+
+        if (bodyFile) {
+            NSData *data;
+            int i = 0;
+            do {
+                data = [bodyFile readDataOfLength:512*1024];
+                [client writeData:data];
+                i += data.length;
+                NSLog(@"%@ %d", url, i);
+            } while (data.length);
+        }
+
+        [request endAccessingResources];
+        [self endRequest:client];
+    }];
+}
+
+- (void)endRequest:(NSFileHandle *)client {
+    [NSNotificationCenter.defaultCenter removeObserver:self name:NSFileHandleDataAvailableNotification object:client];
+    CFDictionaryRemoveValue(self.messages, (__bridge CFTypeRef) client);
+    [client closeFile];
+}
+
+@end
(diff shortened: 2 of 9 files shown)


app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+    }
+    int one = 1;
+    if (setsockopt(CFSocketGetNative(socket), SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one)) != 0) {
+        NSLog(@"APK server: failed to set SO_REUSEADDR");
+        return;
+    }
+    struct sockaddr_in sin = {
+        .sin_len = sizeof(sin),
+        .sin_family = AF_INET,
+        .sin_port = htons(42069),
```
saagarjha, line 45:
```suggestion
        .sin_port = htons(42069), // highly nice
```

app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+        NSLog(@"APK server: failed to bind");
+        return;
+    }
+    self.server = [[NSFileHandle alloc] initWithFileDescriptor:CFSocketGetNative(socket) closeOnDealloc:YES];
+    [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(onAccept:) name:NSFileHandleConnectionAcceptedNotification object:self.server];
+    [self.server acceptConnectionInBackgroundAndNotify];
+    NSLog(@"APK server: serving");
+}
+
+- (void)onAccept:(NSNotification *)n {
```
saagarjha, line 59: `n`, really?

app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+        return [self end];
+    if (CFHTTPMessageIsHeaderComplete(self.message))
+        return [self sendResponse];
+    [self.client waitForDataInBackgroundAndNotify];
+}
+
+- (void)sendResponse {
+    NSURL *url = CFBridgingRelease(CFHTTPMessageCopyRequestURL(self.message));
+    NSArray<NSString *> *components = url.pathComponents;
+    if ([components[0] isEqualToString:@"/"])
```
saagarjha, line 101: `[components firstObject]`?

app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+            status = 500;
+            body = [error.description dataUsingEncoding:NSUTF8StringEncoding];
+        } else {
+            NSLog(@"APK server: serving %@", resourceTag);
+            // TODO: this loads the entire file in memory which is wasteful and unnecessary
+            body = [NSData dataWithContentsOfURL:[request.bundle URLForResource:resourceTag withExtension:nil]];
+        }
+
+        CFHTTPMessageRef response = CFHTTPMessageCreateResponse(kCFAllocatorDefault, status, NULL, kCFHTTPVersion1_1);
+        CFHTTPMessageSetBody(response, CFBridgingRetain(body));
```
saagarjha, line 123: Why is the retain necessary?

app/AppDelegate.m
```diff
@@ -228,7 +230,10 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     };
     SCNetworkReachabilitySetCallback(self.reachability, NetworkReachabilityCallback, &context);
     SCNetworkReachabilityScheduleWithRunLoop(self.reachability, CFRunLoopGetMain(), kCFRunLoopCommonModes);
-    
+
+    self.apkServer = [APKServer new];
```
saagarjha, line 234: `[self.apkServer = [APKServer new] start]`

deps/aports/download-repos.py
```diff
@@ -0,0 +1,58 @@
...
+    if path.exists(): return
+    try:
+        urllib.request.urlretrieve(url, path)
+    except Exception as e:
+        print('failed to download', url, ':', e)
+        raise
+    return pkg
+
+if __name__ == '__main__':
+    download_repo('https://f001.backblazeb2.com/file/alpine-archive', 'main/x86', 'APKINDEX-v3.12-2020-11-15.tar.gz')
```
saagarjha, line 58: Can these be parameters or something that is automatically up-to-date

saagarjha (commented):
Preliminary comments

app/AppDelegate.m
tbodt: aaaaa

genaro120 (approved):

deps/aports/download-repos.py
tbodt: This is supposed to require manual updating so builds are reproducible

app/APKServer.m
tbodt: That would explain the memory leak

app/APKServer.h
```diff
@@ -0,0 +1,18 @@
...
+//  Created by Theodore Dubois on 11/21/20.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface APKServer : NSObject
+
+- (void)start;
```
saagarjha, line 14: This is the saddest class ever. Should it be possible to stop the `APKServer`? If not, is it reasonable to create a server and not start it? If not, shouldn't it just start the moment you initialize it?

app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+
+@interface APKRequest : NSObject
+
+@property NSFileHandle *client;
+@property CFHTTPMessageRef message;
+- (instancetype)initWithClient:(NSFileHandle *)client;
+
+@end
+
+static NSMutableArray<APKRequest *> *extantRequests;
```
saagarjha, line 27: Why `static` rather than a property on the server?

app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+    if (socket == nil) {
+        NSLog(@"APK server: failed to create socket");
+        return;
+    }
+    int one = 1;
+    if (setsockopt(CFSocketGetNative(socket), SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one)) != 0) {
+        NSLog(@"APK server: failed to set SO_REUSEADDR");
+        return;
+    }
+    struct sockaddr_in sin = {
```
saagarjha, line 42: Someone's not getting presents for Christmas

app/APKServer.m
```diff
@@ -0,0 +1,141 @@
...
+    [self.server acceptConnectionInBackgroundAndNotify];
+}
+
+- (void)onReadable:(NSNotification *)n {
+
+}
+
++ (void)initialize {
+    extantRequests = [NSMutableArray new];
+}
```
saagarjha, line 71: Plus this yuckiness would go away too

app/APKServer.m
saagarjha: `(__bridge)`

deps/aports/download-repos.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+import tqdm
```
saagarjha, line 2: do we really need dependencies

deps/aports/download-repos.py
```diff
@@ -0,0 +1,58 @@
...
+    urllib.request.urlretrieve(index_url, index_path)
+    with tarfile.open(index_path) as tar:
+        pkgs = read_index(tar.extractfile('APKINDEX'))
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=30) as pool:
+        futures = [pool.submit(download_package, repo, f'{root_url}/{repo_name}', pkg) for pkg in pkgs]
+        for f in tqdm.tqdm(concurrent.futures.as_completed(futures), total=len(futures)):
+            f.result()
+
+def download_package(repo, root_url, pkg):
```
saagarjha, line 45: throttling is a thing you should do

deps/aports/download-repos.py
```diff
@@ -0,0 +1,58 @@
...
+import urllib.request
+import urllib.parse
+import tarfile
+import pathlib
+import concurrent.futures
+
+IX_NAME = 'P'
+IX_VERSION = 'V'
+
+def read_index(index):
```
saagarjha, line 12: `pathlib.Path` exists

deps/aports/odrs.py
```diff
@@ -0,0 +1,103 @@
+import datetime
+import pathlib
+import plistlib
+import os
+import shutil
+import subprocess
+import ctypes
+
+if os.environ['ENABLE_ON_DEMAND_RESOURCES'] == 'YES':
+    raise Exception('This script collides with Xcode\'s ODR processing')
```
saagarjha, line 10: This is why we use double quotes

deps/aports/odrs.py
```diff
@@ -0,0 +1,103 @@
...
+        pack_path = odrs_dir/(pack_id+'.assetpack')
+        if pack_path.exists():
+            shutil.rmtree(pack_path)
+        pack_path.mkdir()
+        total_size = 0
+        with (pack_path/'Info.plist').open('wb') as f:
+            plistlib.dump({
+                'CFBundleIdentifier': pack_id,
+                'Tags': tags,
+            }, f, fmt=plistlib.FMT_BINARY)
```
saagarjha, line 58: Would it be possible to use the XML format

deps/aports/update-archive.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -x
+# This script only works on my machine
+function sync_repo() {
+    version="$1"
+    path="$2"
+    rclone copy -v --transfers=32 "alpine:$version/$path" "b2:alpine-archive/$path"
+    date=$(date +%F)
+    rclone moveto "b2:alpine-archive/$path/APKINDEX.tar.gz" "b2:alpine-archive/$path/APKINDEX-$version-$date.tar.gz"
+}
+sync_repo v3.12 main/x86
```
saagarjha, line 10: Can we not hardcode this version?

saagarjha (commented):
More stuff

app/APKServer.m
tbodt: I would rather it not exist, perhaps instead I can have APKRequest have a strong reference to itself and clear it to free itself

deps/aports/update-archive.sh
tbodt: Sure, though no one else is supposed to run this script

deps/aports/odrs.py
tbodt: Is it better?

app/APKServer.m
saagarjha: A server keeping track of its requests seems like a fairly straightforward design to me. Why do you think creating a reference cycle is a better idea?

deps/aports/odrs.py
saagarjha: It's readable

tbodt: Looks like this file gets shipped in the app store as-is, so we should make it as small as possible. It's 300k in binary form, and about 4x that in xml form.

app/APKServer.h
tbodt: Will make it start on creation

app/APKServer.m
tbodt: Refactored

deps/aports/download-repos.py
tbodt: How does that help here?

deps/aports/download-repos.py
tbodt: Why

app/APKServer.m
tbodt: Yes

app/APKServer.m
```diff
@@ -0,0 +1,139 @@
...
+
+- (void)onReadable:(NSNotification *)n {
+    NSFileHandle *client = n.object;
+    CFHTTPMessageRef message = (CFHTTPMessageRef) CFDictionaryGetValue(self.messages, (__bridge CFTypeRef) client);
+    NSData *data = [client availableData];
+    if (data.length == 0)
+        return [self endRequest:client];
+    if (!CFHTTPMessageAppendBytes(message, data.bytes, data.length))
+        return [self endRequest:client];
```
saagarjha, line 73: Should these be combined?

app/APKServer.m
```diff
@@ -0,0 +1,139 @@
...
+        NSData *body = nil;
+        NSFileHandle *bodyFile = nil;
+        NSUInteger contentLength = 0;
+
+        if (error == nil) {
+            NSLog(@"APK server: serving %@", resourceTag);
+            bodyFile = [NSFileHandle fileHandleForReadingFromURL:[request.bundle URLForResource:resourceTag withExtension:nil] error:&error];
+            [bodyFile seekToEndOfFile];
+            contentLength = bodyFile.offsetInFile;
+            [bodyFile seekToFileOffset:0];
```
saagarjha, line 100: Is there no better way to find the file size than this?

deps/aports/download-repos.py
```diff
@@ -0,0 +1,68 @@
...
+        pkg_file = f'{pkg[IX_NAME]}-{pkg[IX_VERSION]}.apk'
+        url = f'{root_url}/{repo_name}/{urllib.parse.quote(pkg_file)}'
+        path = repo/pkg_file
+        if path.exists(): continue
+        downloads.append((url, path))
+    download_many(downloads)
+
+def download_one(url, path):
+    n = 0
+    while True:
```
saagarjha, line 50: Can we not use a `while True`

app/APKServer.m
tbodt: Doesn't matter that much

app/APKServer.m
tbodt: NSFileManager is surprisingly bad at this

deps/aports/download-repos.py
tbodt: How? This is the code that parses APKINDEX.tar.gz

deps/aports/download-repos.py
tbodt: The code is cleaner this way, unless you have a suggestion?

This doesn't work, because the exception is swallowed:
```
    for _ in range(3):
        try:
            return urllib.request.urlretrieve(url, path)
        except Exception as e:
            print('retrying', url, e)
```

deps/aports/download-repos.py
saagarjha: It's polite to the servers, since it seems like you're firing off thousands of requests.

tbodt: > Backblaze recommends that you do lots of transfers simultaneously for maximum speed.
https://rclone.org/b2/

saagarjha: "please DDOS me"…well, if that's what they want

deps/aports/download-repos.py
saagarjha: How about something like this (untested):

```python
def read_index(index):
	index = index.read()
	packages = []
	for block in index.split("\n\n"):
		package = {}
		for line in block.split("\n"):
			key, _, value = line.partition(":")
			package[key] = value
		packages.append(package)
	return packages
```

deps/aports/download-repos.py
saagarjha: I would personally store the exception in the `except` block, since I generally don't use `while True`. But it's your choice.

tbodt: I think you lose the stack trace if you try that though

deps/aports/download-repos.py
tbodt: It's only 30 requests at once

PR #1089, opened 2020-11-27
Update README_ZH.md

lolosssss:


tbodt:
Looks like I missed this PR and since then someone else did it

PR #1101, opened 2020-12-06
Added /proc/cpuinfo, /proc/filesystems and enhanced /proc/meminfo.  A…

emkey1:
…lso added Amber appearance.


emkey1:
Will resubmit shortly with some fixes.

PR #1122, opened 2020-12-13
Implement some vec instructions with NEON

nimelehin:
Test results (less is faster):

 CPU | vec_add_b128(ms) | vec_add_b128_neon(ms) | diff | vec_min_ub128(ms) | vec_min_ub128_neon(ms) | diff | vec_compare_eqb128(ms) | vec_compare_eqb128_neon(ms) | diff
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
A14 | 40096,8 | 36707,0 | +9,2% | 63393,6 | 36696,8 | 72,7 | 40037,8 | 36717,0 | +9,04%
A13 | 60300,0 | 41331,4 | +45,9% | 71451,8 | 41417,6 | 72,5 | 60226,2 | 41406,8 | +45,5%
A12x | 65126,2 | 44755,6 | +45,5% | 78699,4 | 45598,2 | 72,6 | 66397,8 | 45160,4 | +47,0%


tbodt:
What effect does this have on non-micro benchmarks?

nimelehin:
> What effect does this have on non-micro benchmarks?

As a test I ran a simple C program working with sse, the performance is about 10-15%, but that seems to be a micro-bench too.
Tried to run GO, it uses sse actively, but no performance difference is seen in go1 test suite, according to profiler bottlenecks in go1 are not sse emulation.

filfat:
v8 uses sse a fair bit. How’s the performance there @nimelehin?

PR #1231, opened 2021-01-07
New App Icon Set

jackvz:
Hi there. This is great! I quickly updated the app icon to a free console icon, https://icons8.com/icon/79750/console. Why the dollar sign as the icon?


Mnpn03:
Hi! There's an open issue where you can submit cool icons: #578. (You can select an app icon from the settings page in the app.)
Also, the icon images in this PR seem to contain a lot of transparency, not quite sure how well this would work as an app icon in iOS (see [this section](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon) of the Apple Interface Guidelines).

The dollar sign is typically associated with shells.

tbodt:
I won't change the default icon, but you can submit icons over at #578.

I would probably not use this one, as it's intended to be used in an icon font at small sizes rather than an app store icon.

jackvz:
You can select an app icon in the app's settings? Okay, this is REALLY great! Yep, should have checked the actual icon quality before doing a PR.

PR #1305, opened 2021-02-28
Accepted relicense

0b101:
I hereby accept the relicence of [iSH](https://github.com/ish-app/ish) from GPLv3 to GPLv2.

This should be the name and email used for my git contributions. If not, I can change it accordingly.


tbodt:
Looks good. Thanks!

PR #1311, opened 2021-03-01
GPLv2: Christopher Albert

krystophny:
This is to confirm that I agree with relicensing my contributions to iSH under the GPLv2.


tbodt:
Thanks!

PR #1379, opened 2021-04-28
Use xcconfig for Development Team ID

rinsuki:


tbodt:
The one problem with this is Xcode can't autocomplete the team ID if it's in an xcconfig. I do like being able to set it in one place though. Maybe the team id could be inherited from the project settings?

rinsuki:
I want to use xcconfig because if I write my team ID to project.pbxproj, I may miss the .pbxproj file diff and commit to git repository.

> The one problem with this is Xcode can't autocomplete the team ID if it's in an xcconfig.

I think too, but autocomplete needs write Team ID directly to .pbxproj (and that makes diff in .pbxproj file)...

maybe we can solve this by add "If Team ID written in pbxproj file directly, replace with `$(ISH_DEVELOPMENT_TEAM)` and write actual Team ID to xcconfig" script to first of build phase or something...? (but i think these script is complicated)

SBCutz111:
McGrew

saagarjha:
Hi, we forgot all about this pull request and ended up implementing essentially the same thing ourselves: https://github.com/ish-app/ish/commit/33642f2ec7d3ed99521e1fb44e48284290b230de. I guess you can get honorary credit for getting this done :)

SBCutz111:
Oleg

PR #1429, opened 2021-05-31
Unaligned atomic cmpxchg8b

kmeisthax:
This PR adds an unaligned mode to our implementation of `lock cmpxchg8b` that uses AArch64's `ldaxp` instruction to do 128-bit *aligned* loads and then manually align the loaded bits with shifts.

This isn't perfect; we make no attempt to check if we are off the end of the unaligned load. Technically speaking, we cannot guarantee atomicity for cross-cache-line loads... but then again, neither did Intel. Badly misaligned loads get broken into two nonatomic loads, which matches all of the references I can find about x86.

This PR fixes MySQL crashing the emulator. It now throws up an error about Linux Direct AIO being unimplemented.

### DIFF
--- a/jit/gadgets-aarch64/misc.S
+++ b/jit/gadgets-aarch64/misc.S
@@ -121,9 +121,9 @@
     write_bullshit 64, atomic_cmpxchg8b
 
 2: #UNaligned cmpxchg8b (emulated with aligned 16b loads)
-    and x17, _xaddr, 7
+    and x17, _xaddr, 0xF
     lsl x17, x17, 3
-    and _xaddr, _xaddr, -7
+    and _xaddr, _xaddr, -15
     
     write_prep 128, atomic_cmpxchg8b_unaligned
     mov w10, eax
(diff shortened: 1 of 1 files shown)


kmeisthax:
I added a codepath for *really* badly aligned addresses. It sacrifices atomicity, but [as far as I can tell](https://stackoverflow.com/questions/1415256/alignment-requirements-for-atomic-x86-instructions-vs-mss-interlockedcompareex) x86 processors do basically the same thing.

kmeisthax:
@tbodt Is there any chance I could get you or one of the other maintainers to review this PR? AFAIK it's complete, it just needs CI approval and a review.

tbodt:
Will go ahead and approve for CI. I'll have to let someone more familiar with arm64 to review.

kmeisthax:
A few months ago on Discord we talked about this PR and came to the conclusion that using `ldaxp` or dropping atomicity on unaligned loads is not the right approach. I'm withdrawing this PR and filing a new one that only issues segfaults to the guest process (so we don't crash the host iOS process).

PR #1478, opened 2021-06-29
Stub missing syscall 364, also added comment for syscall 383

### LINKED ISSUE #1324: Adb error protocol fault
Whenever I run the command adb devices I get thisimage
I just need it for simple app installs so doesn't matter if it doesn't work fully

guanzhangrtk:
This fixes #1324 on master which is currently running alpinelinux v3.14


tbodt:
Neat!

PR #1584, opened 2021-10-10
Async I/O (syscalls 245 - 249)

kmeisthax:
* https://man7.org/linux/man-pages/man2/io_setup.2.html
 * https://man7.org/linux/man-pages/man2/io_submit.2.html
 * https://man7.org/linux/man-pages/man2/io_cancel.2.html
 * https://man7.org/linux/man-pages/man2/io_getevents.2.html
 * https://man7.org/linux/man-pages/man2/io_destroy.2.html

Linux has an async I/O facility (separate from epoll) that allows sending a small handful of file operations (notably, positioned and/or vectored reads and writes, syncs, and possibly also poll) on an already-open FD into an in-kernel context, and then retrieving all of their results at a later time. It's almost like an early prototype of modern Linux's `io_uring`, but using syscalls instead of ring buffers.

- [x] Host-side memory structures for AIO contexts and task tables
- [x] `io_setup` impl
- [x] `io_destroy` impl
- [x] `io_submit` impl
  - [x] FD method for IO submit
  - [x] Fallback `io_submit` impl
    - [x] `IOCB_CMD_PREAD`
    - [x] `IOCB_CMD_PWRITE`
    - [x] `IOCB_CMD_FSYNC`
    - [x] `IOCB_CMD_FDSYNC`
    - ~~`IOCB_CMD_POLL`~~ I don't think this will be possible
    - [x] `IOCB_CMD_NOOP`
    - [x] `IOCB_CMD_PREADV`
    - [x] `IOCB_CMD_PWRITEV`
  - ~~Non-blocking fallback, leveraging `O_NONBLOCK` and poll/epoll where available~~ Neither will this.
- [x] `io_getevents` impl
  - [x] Basic impl
  - [x] `min_nr` and `timeout` support
- ~~`io_cancel` impl~~ I actually can't figure out how this is supposed to work
- [x] e2e tests
  - [x] Figure out why the tests expect `evt.obj` to be sign-extended


kmeisthax:
I've managed to get this to build again. I'm going to release the draft flag on this PR. It should be sufficient enough to run software that needs async I/O, albeit not with the speed benefits. A good idea for a future PR would be hooking this system up to POSIX AIO (or whatever else iOS supports) to do proper parallel IO on the host filesystem.

I wasn't able to write tests for `io_cancel`; and all the software I want to run on iSH doesn't use it. If anyone has a compelling use case for I/O cancellation let me know; otherwise I'm going to leave it as `ENOSYS`.

kernel/aio.c
```diff
@@ -0,0 +1,448 @@
...
+#include "kernel/task.h"
+#include "kernel/aio.h"
+#include "kernel/fs.h"
+#include "kernel/time.h"
+#include "fs/aio.h"
+#include "fs/fd.h"
+
+// Guest memory offsets for the IOCB structure.
+// Calculated by a test program compiled and ran in iSH itself.
+const size_t AIO_IOCB_DATA = 0;
```
tbodt, line 12: Why not define a struct, instead of reading and writing individual offsets? Then these could become static_assert lines to verify that the struct layout is correct.

kernel/task.h
```diff
@@ -36,6 +37,9 @@ struct task {
     struct fdtable *files;
     struct fs_info *fs;
 
+    // Currently active AIO contexts. Contains internal lock.
+    struct aioctx_table *aioctx;
```
tbodt, line 41: This seems like it would make more sense as an inline struct instead of a pointer to a malloced block, since it always exists for exactly as long as the task.

fs/aio.c
```diff
@@ -0,0 +1,309 @@
...
+    if (ctx == NULL) return _EINVAL;
+
+    lock(&ctx->lock);
+    int err = wait_for(&ctx->cond, &ctx->lock, timeout);
+    unlock(&ctx->lock);
+
+    return err;
+}
+
+void aioctx_lock(struct aioctx* ctx) {
```
tbodt, line 187: consistency nit
```suggestion
void aioctx_lock(struct aioctx *ctx) {
```

fs/aio.c
```diff
@@ -0,0 +1,309 @@
...
+    if (ctx == NULL) return;
+
+    lock(&ctx->lock);
+}
+
+void aioctx_unlock(struct aioctx* ctx) {
```
tbodt, line 193:
```suggestion
void aioctx_unlock(struct aioctx *ctx) {
```

tbodt (commented):
Neat! Did a quick review now, soon I hope to find the time to learn a bit about how AIO works so I can do a real review.

PS: please fix the tests :)

rtsui-harmonicinc:
Would this PR still be merged?  Packages like Neovim on Newer alpine versions like 3.24 depends on this

PR #1664, opened 2021-12-25
Add /proc/ish

saagarjha:
Create /proc/ish and populate /proc/ish/version to start with.

### DIFF
--- a/app/AppDelegate.m
+++ b/app/AppDelegate.m
@@ -35,7 +35,6 @@
 @interface AppDelegate ()
 
 @property BOOL exiting;
-@property NSString *unameVersion;
 @property NSString *ishVersion;
 @property NSString *unameHostname;
 @property SCNetworkReachabilityRef reachability;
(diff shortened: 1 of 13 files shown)


fs/proc/ish.h
```diff
@@ -0,0 +1,15 @@
+#ifndef FS_PROC_ISH_H
+#define FS_PROC_ISH_H
+
+int proc_ish_show_version(struct proc_entry *UNUSED(entry), struct proc_data *buf);
+
+#define PROCFS_ISH_ENTRIES \
```
tbodt, line 6: Why a macro? Should just be in ish.c

fs/proc/ish.h
```diff
@@ -0,0 +1,15 @@
...
+    {"version", .show = proc_ish_show_version},
+
+#define PROCFS_ISH_ENTRY_COUNT (sizeof((struct proc_dir_entry []) { PROCFS_ISH_ENTRIES }) / sizeof(struct proc_dir_entry))
```
tbodt, line 9: This sucks

fs/proc/ish.h
```diff
@@ -0,0 +1,15 @@
...
+
+struct proc_dir_entry proc_ish_entries[PROCFS_ISH_ENTRY_COUNT];
+
+#define PROCFS_ISH_ENTRY {"ish", S_IFDIR, .children = proc_ish_entries, .children_sizeof = sizeof(proc_ish_entries)}
```
tbodt, line 13: Same here, though I can somewhat more understand why this macro

saagarjha:
Yeah, all of these suck but they're necessary to make the `proc_ish_entries` be defined with the correct size. I think this is the best option among several poor ones.

PR #1665, opened 2021-12-26
Add /proc/filesystems

emkey1:
Welcome to Alpine!

You can install packages with: apk add <package>

You may change this message by editing /etc/motd.

MacBook-Pro:~# cat /proc/filesystems 
nodev    real
nodev    proc
nodev    devpts
nodev    tmpfs
nodev    ios
nodev    ios-unsafe
MacBook-Pro:~#


emkey1:
I'll try again later.  Sorry!

PR #1666, opened 2021-12-26
Add /proc/filesystems

emkey1:
Welcome to Alpine!

You can install packages with: apk add

You may change this message by editing /etc/motd.

MacBook-Pro:# cat /proc/filesystems
nodev real
nodev proc
nodev devpts
nodev tmpfs
nodev ios
nodev ios-unsafe
MacBook-Pro:#

### DIFF
--- a/fs/mount.c
+++ b/fs/mount.c
@@ -23,6 +23,21 @@ void fs_register(const struct fs_ops *fs) {
     assert(!"reached filesystem limit");
 }
 
+char * get_filesystems(void) {
+    char *fs_list = calloc(MAX_FILESYSTEMS * 50, sizeof(char)); // Reasonable assumption?
+    unsigned int i;
+
+    for ( i = 0; i < MAX_FILESYSTEMS; i++ ) {
+        if (filesystems[i] != NULL) {
+            fs_list = strcat(fs_list, "nodev    ");
+            fs_list = strcat(fs_list, filesystems[i]->name);
+            fs_list = strcat(fs_list, "\n");
+        }
+    }
+
+    return fs_list;
+}
+
 struct mount *mount_find(char *path) {
     assert(path_is_normalized(path));
     lock(&mounts_lock);
--- a/fs/proc/root.c
+++ b/fs/proc/root.c
@@ -22,6 +22,12 @@ static void show_kb(struct proc_data *buf, const char *name, uint64_t value) {
     proc_printf(buf, "%s%8"PRIu64" kB\n", name, value / 1000);
 }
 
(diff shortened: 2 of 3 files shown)


fs/mount.c
```diff
@@ -23,6 +23,21 @@ void fs_register(const struct fs_ops *fs) {
     assert(!"reached filesystem limit");
 }
 
+char * get_filesystems(void) {
+    char *fs_list = calloc(MAX_FILESYSTEMS * 50, sizeof(char)); // Reasonable assumption?
```
saagarjha, line 27: Thanks for sending this PR in! However, there's one problem here: your code allocates this buffer here, but it fails to free it. So each time the filesystems list is requested this leaks some memory. The buffer you're supposed to be writing to is already known to the procfs functions (it uses `proc_printf` which will do resizing as necessary) so I would suggest you actually use it to construct the string instead of creating a buffer here and limiting its size. What you may wish to do instead is return a list of filesystems instead of their string representations, and let the other file handle printing them. Once it's done, it'll free it.

saagarjha (changes requested):

fs/mount.c
```diff
@@ -23,6 +23,21 @@ void fs_register(const struct fs_ops *fs) {
...
+    unsigned int i;
+
+    for ( i = 0; i < MAX_FILESYSTEMS; i++ ) {
```
Mnpn, line 30: nit: spacing, `for (i = 0; i < MAX_FILESYSTEMS; i++) {`

fs/mount.c
emkey1: Got it.  I'll take a look.  Thank you for the feedback.

tbodt:
Looks like this was merged by accident. Reverted it...but github will need a new PR to be created in order to merge it later.

emkey1:
> Looks like this was merged by accident. Reverted it...but github will need a new PR to be created in order to merge it later.

Will do.

fs/mount.c
emkey1: Too much shell script programming?  I'm a fan of white space but that is excessive.  Noted.

YiguiDing:
help！！！
problem again😅：

```bash
iPad:~# service docker start
grep: /proc/filesystems: No such file or directory
grep: /proc/filesystems: No such file or directory
grep: /proc/filesystems: No such file or directory
grep: /proc/filesystems: No such file or directory
grep: /proc/filesystems: No such file or directory
grep: /proc/filesystems: No such file or directory
 * Starting Docker Daemon ...                                                                                                                                [ ok ]
```

PR #1667, opened 2021-12-26
Segfault the guest process when we encounter an unaligned atomic cmpxchg8b

kmeisthax:
Like #1429, but less wrong.

My prior PR misinterpreted certain paragraphs of Intel's processor documentation. Unaligned `lock cmpxchg8b` maintains atomicity *regardless* of alignment; this is not emulatable with ARM atomics alone. Properly emulating this requires a complicated locking setup that would almost certainly harm the performance of well-behaved x86 processes. In fact, I'm pretty sure the prior PR *never worked* - I've noticed this crash doesn't always occur when you launch `mariadbd`; and it's entirely possible that I was just getting very lucky with heap allocations that just so happened to align whatever packed structs are there.

So, instead, here's a new PR that triggers a segfault on unaligned atomic access. This will keep the user from losing any other work they might have been doing in other guest processes.

### DIFF
--- a/jit/gadgets-aarch64/memory.S
+++ b/jit/gadgets-aarch64/memory.S
@@ -83,6 +83,7 @@ NAME(si_gadgets):
         mov _xaddr, x19
         ret
 
+.global segfault_\type
 segfault_\type:
     ldr _addr, [_tlb, -TLB_entries+TLB_segfault_addr]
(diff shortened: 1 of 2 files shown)


tbodt:
LGTM, thanks!

Issue #1237, opened 2021-01-11
Rename "Disable Screen Dimming" to “Keep screen turned on”


StefanLobbenmeier:
#1212: That would make it a lot clearer, I actually did not understand what the option was going to do before I searched for it

andrewintw:
agree

PR #1680, opened 2022-01-02
Rename screen dimming label

Mnpn:
Now says "Keep Screen Turned On" instead of "Disable Screen Dimming", resolving #1237.


saagarjha (approved):
This option happens to be the last one in its section, so if you feel like it could be improved by a short caption describing what it does feel free to add one.

Mnpn:
I like that idea, thanks!

PR #1681, opened 2022-01-02
Expose user defaults via /proc/ish

saagarjha:
High level design:

* User defaults are now available via /proc/ish/defaults and /proc/ish/.defaults. The latter is "all" preferences and the former symlinks into it for "friendly" names and to indicate that these are preferences we explicitly support you changing.
* Removing the file, as well as reading and writing to it is supported. This is roughly equivalent to reseting the preference, getting it, and setting it.
* Preferences are exposed via JSON fragments.

### DIFF
--- a/app/AboutAppearanceViewController.m
+++ b/app/AboutAppearanceViewController.m
@@ -22,8 +22,10 @@
     [super viewDidLoad];
     [UserPreferences.shared observe:@[@"theme", @"fontSize", @"fontFamily", @"hideStatusBar"]
                             options:0 owner:self usingBlock:^(typeof(self) self) {
-        [self.tableView reloadData];
-        [self setNeedsStatusBarAppearanceUpdate];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [self.tableView reloadData];
+            [self setNeedsStatusBarAppearanceUpdate];
+        });
     }];
 }
 
--- a/app/AboutExternalKeyboardViewController.m
+++ b/app/AboutExternalKeyboardViewController.m
@@ -26,7 +26,9 @@ const int kCapsLockMappingSection = 0;
     [super viewDidLoad];
     [UserPreferences.shared observe:@[@"capsLockMapping", @"optionMapping"]
                             options:0 owner:self usingBlock:^(typeof(self) self) {
-        [self.tableView reloadData];
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [self.tableView reloadData];
+        });
     }];
     [self _update];
 }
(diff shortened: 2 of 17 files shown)


emkey1:
I like the PR, but I get SIGABRT when I compile and run iSH with it...

Screen Shot 2022-01-02 at 4 07 26 PM
.

Commenting out the assert gets iSH to run but is probably not the best solution. :)

emkey1:
I'm using a fairly recent version of my root fs, but otherwise my environment is totally vanilla.

saagarjha:
How did you get this to happen? Just running ps?

emkey1:
> How did you get this to happen? Just running ps?

Nope, just starting up the filesystem.  It does do some additional stuff tat the default root fs doesn't.  Starting OpenRC for instance along with dcron.  I'll get a complete list of stuff it does up shortly.

emkey1:
Based on the crash of course, yes it does run ps as part of the startup.  Probably as part of the OpenRC stuff starting up.  Nothing else exotic that I can see.  It uses Alpine 3.12.1.  I'll try my next root FS, which is going to use 3.12.9

saagarjha:
Can you share the rootfs?

emkey1:
> Can you share the rootfs?

Yes, I'll append a link shortly.

emkey1:
> Can you share the rootfs?

It can be found here...

https://discord.com/channels/776432683302649866/778618184919285810

I just verified that the .90 build crashes with this PR installed on vanilla iSH.  

sudo was the offending command, but that was likely because there are extra things that happen the first time the image is booted.  Mostly setting up OpenRC.  /etc/profile is where this happens.

saagarjha:
I don't think I have access to that file :(

emkey1:
> I don't think I have access to that file :(

Hmm, that should just be a Discord channel from my Discord.  Let me see if I can get a direct link...

https://cdn.discordapp.com/attachments/778618184919285810/927384487845851228/ALPINE_3.12.9-iSH-AOK_ALPHA_90.tgz

Does that work?

saagarjha:
It does, and I can reproduce. Thanks!

saagarjha:
Turns out I had a typo:

```diff
diff --git a/fs/proc.c b/fs/proc.c
index 4ce2092..23c6165 100644
--- a/fs/proc.c
+++ b/fs/proc.c
@@ -150,7 +150,7 @@ static ssize_t proc_pread(struct fd *fd, void *buf, size_t bufsize, off_t off) {
 }

 static void proc_buf_write(struct proc_data *buf, const void *data, size_t size, size_t off) {
-    assert(off <= size);
+    assert(off <= buf->size);
     size_t tidemark = off + size;
     if (tidemark > buf->capacity) {
         size_t new_capacity = buf->capacity + 1;
```

Issue #1704, opened 2022-01-28
Create SECURITY.md


JamieSlome:
Hey there!

I belong to an open source security research community, and a member (@michaellrowley) has found an issue, but doesn’t know the best way to disclose it.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

tbodt:
Feel free to contact me via the email on my github profile. Please understand though that I don't really want to maintain a security boundary between iSH and the iOS sandbox. (Maybe it would be fun to add a /dev/mem for arbitrary memory access by design!)

JamieSlome:
@tbodt - thanks for your response and update here! ❤️ 

If you would prefer, you can directly access the report here:
https://huntr.dev/bounties/43301686-ef08-40eb-8d86-1d50aeb17f81/

It is private and only accessible to maintainers with repository write permissions.

tbodt:
We now have a SECURITY.md: https://github.com/ish-app/ish/blob/master/SECURITY.md

PR #1705, opened 2022-01-29
Add SECURITY.md

Rhinozz:
Upon the request of JamieSlome (#1704). This can be expanded if need be - right now it just contains tbodt's email address, as [recommended](https://github.com/ish-app/ish/issues/1704#issuecomment-1024727717) by him.


tbodt:
Thanks for the PR, but we're currently discussing what we want our SECURITY.md to be, so we don't really need this for now.

Issue #1699, opened 2022-01-24
Cannot open /dev/fd/63: no such file or directory


xezo360hye:
Using structure `command1 <(command2)` results in the error from title. Tried workarounds like `ln -s /proc/self/fd /dev/fd` or installing `udev` but this didn't help - _/proc/self/fd_ contains only _1, 2, 3 and 4_ files. Any suggestions if it's fixable locally?

robCam:
This is somewhat analogous to issue [#164](https://github.com/ish-app/ish/issues/164).

However, running the [example script](https://github.com/ish-app/ish/issues/164#issue-387330888) from issue `#104` results in the following error:

> ```
> line 5: /dev/fd/63: No such file or directory
>                 ^^
>```

**Note:** I get `/dev/fd/63` and not `/dev/fd/62` as reported in issue `#104`.

xezo360hye:
> This is somewhat analogous to issue [#164](https://github.com/ish-app/ish/issues/164).
Yeah, it is, I didn’t see that issue because of its title

> **Note:** I get `/dev/fd/63` and not `/dev/fd/62` as reported in issue `#104`.
It vary sometimes, earlier I had the same problem and it also pointed to /dev/fd/62

kalebmckale:
Got `cat: can't open '/dev/fd/63': No such file or directory` when trying to use pyenv. Any updates on a fix?

kalebmckale:
Ideas?

ArshadSiddiqui:
I get the same error while running "pass generate"
```
pass generate tmp 8
/usr/bin/pass: line 538: /dev/fd/63: No such file or directory
Could not generate password from /dev/urandom.
```

Truth1984:
on alpine 

```bash
mkfifo pipe
echo echo hi > pipe &
bash < pipe
```

works fine, but ish stuck on `bash < pipe` part, and sometimes it will even stuck on `> pipe &` section

Issue #783, opened 2020-06-17
Process substitution in zsh


romkatv:
To reproduce, install zsh on iSH and run the following command:

```zsh
zsh -fc 'zmodload zsh/system && sysopen -r -o cloexec -u fd <(true)'
```

Expected: the command succeeds with no output.

Actual: this error:

```text
zsh:sysopen:1: can't open file /proc/self/fd/12: no such file or directory
```

---

I don't use iSH myself. This came my way via https://github.com/romkatv/gitstatus/issues/150. @xinity, please provide additional details if iSH dev(s) ask for them.

tbodt:
Duplicate of #164

Issue #164, opened 2018-12-04
Process substition in BASH fails.


kherge:
## Example Script

```sh
#!/usr/bin/env bash

while read LINE; do
    echo “$LINE”
done < <(echo hello)
```

## Expected Result

    $ hello

## Actual Result

    line 5: /dev/fd/62: No such file or directory

## References

1. http://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html#Process-Substitution
2. https://stackoverflow.com/a/2443120

tbodt:
What's missing now is the very weird (to me) linux behavior where pipe file descriptors appear to be broken symlinks but can actually be opened, and become a second file descriptor for the same pipe.

xinity:
@tbodt any hints to fix this ? as an example we can't leverage powerline10k or any advanced (and very useful features) of ZSH :(

Do you need more tests from us ? 
let me know :)

Keep up the awesome work, discovered your app few days ago , NIFTY !!!!

tbodt:
The problem is well understood (https://github.com/ish-app/ish/issues/164#issuecomment-459837509) but I am not quite sure how to solve it.

f-eric:
This is pretty much over my head but I've been looking around a bit and I stumble upon https://github.com/microsoft/WSL/issues/266 which addresses the same thing(?). I'm not sure if it's helpful, but maybe...

tbodt:
Yeah, WSL had the same problem. Not much info there about how they solved it though.

kherge:
I apologize that this is also over my head, but maybe it's buried in one or more commits for their kernel? I tried to create [a narrow search](https://github.com/search?q=proc+fd+repo%3Amicrosoft%2FWSL2-Linux-Kernel+created%3A%3E2016-04-25+created%3A%3C2016-06-08&type=Commits&ref=advsearch&l=&l=) that could hopefully be a starting point.

tbodt:
The bug report you linked is for WSL 1, which is not open source. This was never a problem in WSL 2.

f-eric:
I wish I was a developer... For what it's worth I asked in the WSL issue if there would be any possibility for them to share how it was solved. I'm not sure how I'm able to help in any other way. The worst that can happen is no one answers or we get a no.

makakin:
Any solution to fix this issue?

le0-VV:
Is this thread still going? Came across this problem and is now mildly infuriating me

kamikaze3384:
[...]

PR #1716, opened 2022-02-10
Support substitution

luyui:
Should workaround #164 #783  #1699 etc.
Requires a manual creation of:
```shell
$ ln -s /proc/self/fd /dev/fd
```

### DIFF
--- a/fs/adhoc.c
+++ b/fs/adhoc.c
@@ -17,6 +17,34 @@ struct fd *adhoc_fd_create(const struct fd_ops *ops) {
     return fd;
 }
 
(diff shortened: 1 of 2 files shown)


kalebmckale:
Did this get merged?

e-maryniak:
Any idea when this will get merged?

Just as a side note: bash does not run, by default, in Posix mode.
Posix mode does not support bash's process substitution "cmd <( ... )".
You can check this with 'set -o' (it should list "posix off").
But indeed iSH is not running in Posix mode, so that did not explain why process substitution did not work.
So please accept this merge request, as I'm using process substitution frequently in my bash scripts!

PR #1740, opened 2022-02-27
Custom theme support

saagarjha:
Most of the stuff you'd want: system appearance matching, default themes, custom user-specified themes, procfs and documents directory integration.

### DIFF
--- a/app/TerminalView.m
+++ b/app/TerminalView.m
@@ -71,7 +71,7 @@ struct rowcol {
             self->_keyCommands = nil;
         });
     }];
-    [prefs observe:@[@"fontFamily", @"fontSize", @"theme"]
+    [prefs observe:@[@"colorScheme", @"fontFamily", @"fontSize", @"theme", @"cursorStyle", @"blinkCursor"]
            options:0 owner:self usingBlock:^(typeof(self) self) {
         dispatch_async(dispatch_get_main_queue(), ^{
             [self _updateStyle];
@@ -154,13 +154,6 @@ static NSString *const HANDLERS[] = {@"syncFocus", @"focus", @"newScrollHeight",
 
 #pragma mark Styling
 
-- (NSString *)cssColor:(UIColor *)color {
-    CGFloat red, green, blue, alpha;
-    [color getRed:&red green:&green blue:&blue alpha:&alpha];
-    return [NSString stringWithFormat:@"rgba(%ld, %ld, %ld, %ld)",
-            lround(red * 255), lround(green * 255), lround(blue * 255), lround(alpha * 255)];
-}
-
 - (void)_updateStyle {
     NSAssert(NSThread.isMainThread, @"This method needs to be called on the main thread");
     if (!self.terminal.loaded)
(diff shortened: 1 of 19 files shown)


Serena-io (approved):

tbodt (commented):
Found a couple of bugs playing around:
- I swiped to delete a theme, the theme didn't go away, I swiped a little less and tapped the delete button, and then I got an array index crash because the theme was actually deleted the first time but the table view didn't update.
- If you tap edit, then tap a theme, then go back, the table view is no longer in editing mode (it should still be).

Comments on UI:
- Instead of having to teach people to swipe to find the duplicate button, let's put the duplicate button inside the editing screen. Also add a new theme button on the main page that duplicates the current theme.
- The dynamic type is correct on the themes page, which is jarring because it's not correct anywhere else in the settings screen. We should probably fix everything else instead of fixing this though.

fs/proc/ish.c
```diff
@@ -14,6 +14,44 @@ bool (*get_user_default)(const char *name, char **buffer, size_t *size);
 bool (*set_user_default)(const char *name, char *buffer, size_t size);
 bool (*remove_user_default)(const char *name);
 
+static int proc_ish_show_colors(struct proc_entry *UNUSED(entry), struct proc_data *buf) {
+    proc_printf(buf,
+                "\x1B[30m" "iSH" "\x1B[39m "
```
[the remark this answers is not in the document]
saagarjha: They're actually different; this one has carriage returns to make it work when piped into a raw tty. Plus framing to make it look pretty. It sucks to have both but I considered the complexity associated with actually running the command and decided it wasn't worth it here.

saagarjha:
Ok, I've pushed some new changes. This should resolve the crashes when editing the theme list and the desynced "Edit" button. I've also added a "Duplicate" bar button item for each theme. However, on the main page I chose instead to have a new item appear, "Import theme", when you enter editing mode. I feel this is more useful, especially since we expose the "Duplicate" action in two other places.

app/TerminalView.m
```diff
@@ -168,12 +168,21 @@ - (void)_updateStyle {
...
         _overrideFontSize = 0;
-    id themeInfo = @{
+    Palette *palette = prefs.palette;
+    if (self.overrideAppearance != OverrideAppearanceNone) {
+        palette = self.overrideAppearance == OverrideAppearanceLight ? prefs.theme.lightPalette : prefs.theme.darkPalette;
+    }
+    NSMutableDictionary<NSString *, id> *themeInfo = [@{
         @"fontFamily": prefs.fontFamily,
         @"fontSize": @(self.effectiveFontSize),
-        @"foregroundColor": [self cssColor:prefs.theme.foregroundColor],
```
tbodt, line 174: Can delete cssColor: now

app/TerminalViewController.m
```diff
@@ -368,6 +368,16 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
     }
 }
 
+- (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
```
tbodt, line 371: Why not add a traitCollectionDidChange to the various views and view controllers that need appearance updates?

app/Theme.m
```diff
@@ -0,0 +1,358 @@
...
+    return strdup(NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES).firstObject.UTF8String);
+}
+
+@implementation UIColor (iSH)
+- (nullable instancetype)ish_initWithHexString:(NSString *)string {
+    if (!string.length) {
+        return nil;
+    }
+    NSScanner *scanner = [NSScanner scannerWithString:string];
+    // Skip the leading #
```
tbodt, line 22: check that it actually starts with a #?

app/UserPreferences.m
```diff
@@ -184,9 +211,20 @@ - (instancetype)init {
...
+            // This one is a little bit special, so it needs extra handling.
+            // The backing property for this is intentionally underscored.
+            kPreferenceThemeKey: @"userTheme",
         };
 #undef property
+        
+        [self updateTheme];
+        
+        [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(updateTheme:) name:ThemesUpdatedNotification object:nil];
+        [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(updateTheme:) name:ThemeUpdatedNotification object:nil];
```
tbodt, line 227: What's the difference between these two notifications?

app/UserPreferences.m
```diff
@@ -133,20 +146,30 @@ - (instancetype)init {
...
+            kPreferenceBlinkCursorKey: @(NO),
+            kPreferenceCursorStyleKey: @(CursorStyleBlock),
             kPreferenceHideStatusBarKey: @(NO),
+            kPreferenceColorSchemeKey: @(ColorSchemeMatchSystem),
+            kPreferenceThemeKey: @"Default",
         }];
-        _theme = [[Theme alloc] initWithProperties:[_defaults objectForKey:kPreferenceThemeKey]];
+        // https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/
+        if (@available(iOS 13.4, *)) {
+            [_defaults registerDefaults:@{
```
tbodt, line 165: This appears to be redundant with -approximateFont, why not always set this to kSystemMonospacedFontName and let approximateFont change that to menlo as needed?

app/TerminalViewController.m
saagarjha: I picked something that's always in the view hierarchy.

app/UserPreferences.m
saagarjha: The former is when the list of themes changes. The latter is when the current theme changes.

app/UserPreferences.m
saagarjha: `kSystemMonospacedFontName` is morally only supposed to support ui-monospaced, which is the SF Mono font. That -approximateFont returns Menlo if you happen to set this value before 13.4 is really just there as a backstop.

Mnpn (approved):
UI-wise this now LGTM

PR #1777, opened 2022-04-16
Create iOS

joseangelmezavilla:


Mnpn:
Hi! This doesn't look quite right, this pull request only contains a blank file. Did you perhaps mean to open [an issue](https://github.com/ish-app/ish/issues) instead?

PR #1812, opened 2022-05-24
Handle AT_EMPTY_PATH for fstatat64

zhaofengli:
This PR adds correct handling for the `AT_EMPTY_PATH` flag in `fstatat64`. `AT_EMPTY_PATH` allows an empty string to be specified as the path, in which case `fstatat` should operate on the file pointed at by the fd itself:

```
       AT_EMPTY_PATH (since Linux 2.6.39)
              If pathname is an empty string, operate on the file
              referred to by dirfd (which may have been obtained using
              the open(2) O_PATH flag).  In this case, dirfd can refer
              to any type of file, not just a directory, and the
              behavior of fstatat() is similar to that of fstat().  If
              dirfd is AT_FDCWD, the call operates on the current
              working directory.  This flag is Linux-specific; define
              _GNU_SOURCE to obtain its definition.

```

This behavior is relied on by newer glibc when loading shared objects. Currently we return `-ENOENT` for empty paths, leading to behavior like:

```console
/ # bash
bash: error while loading shared libraries: libreadline.so.8: cannot stat shared object: No such file or directory
/ # dmesg
[snip]
2 call 295 openat(-100, "/nix/store/kslb3am4237in6m7i6qi210vc6w7rsfn-readline-8.1p2/lib/libreadline.so.8", 0x88000, 0x0) = 0x3
2 call 3   read(3, 0xffffd110, 512) "ELF" = 0x200
2 call 180 pread(3, 0xffffcee0, 40, 314308) "" = 0x28
2 call 300 stat(at=3, path="", statbuf=0xffffce78, follow_links=1) = 0xfffffffe
2 call 6   close(3) = 0x0
2 call 146 writev(2, 0xffffcd90, 10) {"bash", 4} {": ", 2} {"error while loading shared libraries", 36} {": ", 2} {"libreadline.so.8", 16} {": ", 2} {"cannot stat shared object", 25} {": ", 2} {"No such file or directory", 25} {"
", 1} = 0x73
2 call 252 exit_group(127)
```

With the fix:

```console
/ # bash
bash-5.1# exit
/ # dmesg
[snip]
4 call 295 openat(-100, "/nix/store/kslb3am4237in6m7i6qi210vc6w7rsfn-readline-8.1p2/lib/libreadline.so.8", 0x88000, 0x0) = 0x3
4 call 3   read(3, 0xffffd110, 512) "ELF" = 0x200
4 call 180 pread(3, 0xffffcee0, 40, 314308) "" = 0x28
4 call 300 stat(at=3, path="", statbuf=0xffffce78, flags=0x1000) = 0x0
```

Note that many glibc programs are still broken. I'm playing around with getting a Nix-based environment running in iSH and while I was able to start bash and run some small programs, Nix itself still doesn't run:

- `getcontext()` and `setcontext()` save and restore `fs` (#370)
	- `5 illegal instruction at 0xf749e35f: 66 8c e2 89 50 18 8d 88` ([`mov dx, fs`](https://github.com/bminor/glibc/blob/a01a13601c95f5d111d25557656d09fe661cfc89/sysdeps/unix/sysv/linux/i386/getcontext.S#L42))
- Need to implement/stub more syscalls: 126 (sigprocmask), 435 (clone3)
- ... and more, because Nix still hangs after some quick hacks

### DIFF
--- a/fs/stat.c
+++ b/fs/stat.c
@@ -30,11 +30,31 @@ struct newstat64 stat_convert_newstat64(struct statbuf stat) {
     return newstat;
 }
 
-int generic_statat(struct fd *at, const char *path_raw, struct statbuf *stat, bool follow_links) {
+int generic_statat(struct fd *at, const char *path_raw, struct statbuf *stat, int flags) {
+    int err;
+
+    bool empty_path = flags & AT_EMPTY_PATH_;
+    bool follow_links = !(flags & AT_SYMLINK_NOFOLLOW_);
+
     char path[MAX_PATH];
-    int err = path_normalize(at, path_raw, path, follow_links ? N_SYMLINK_FOLLOW : N_SYMLINK_NOFOLLOW);
-    if (err < 0)
-        return err;
+    if (empty_path && (strcmp(path_raw, "") == 0)) {
+        if (at == AT_PWD) {
+            at = current->fs->pwd;
+        }
+
+        err = generic_getpath(at, path);
+        if (err < 0)
+            return err;
+
+        if (!path_is_normalized(path)) {
+            return _ENOENT;
+        }
+    } else {
+        err = path_normalize(at, path_raw, path, follow_links ? N_SYMLINK_FOLLOW : N_SYMLINK_NOFOLLOW);
+        if (err < 0)
+            return err;
+    }
+
     struct mount *mount = find_mount_and_trim_path(path);
     memset(stat, 0, sizeof(*stat));
     err = mount->fs->stat(mount, path, stat);
@@ -49,17 +69,18 @@ static struct fd *at_fd(fd_t f) {
     return f_get(f);
 }
 
-static dword_t sys_stat_path(fd_t at_f, addr_t path_addr, addr_t statbuf_addr, bool follow_links) {
+// The `flags` parameter accepts AT_ flags
+static dword_t sys_stat_path(fd_t at_f, addr_t path_addr, addr_t statbuf_addr, int flags) {
     int err;
     char path[MAX_PATH];
     if (user_read_string(path_addr, path, sizeof(path)))
         return _EFAULT;
-    STRACE("stat(at=%d, path=\"%s\", statbuf=0x%x, follow_links=%d)", at_f, path, statbuf_addr, follow_links);
+    STRACE("stat(at=%d, path=\"%s\", statbuf=0x%x, flags=0x%x)", at_f, path, statbuf_addr, flags);
     struct fd *at = at_fd(at_f);
     if (at == NULL)
         return _EBADF;
     struct statbuf stat = {};
-    if ((err = generic_statat(at, path, &stat, follow_links)) < 0)
+    if ((err = generic_statat(at, path, &stat, flags)) < 0)
         return err;
     struct newstat64 newstat = stat_convert_newstat64(stat);
     if (user_put(statbuf_addr, newstat))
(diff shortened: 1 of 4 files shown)


zhaofengli:
@emkey1 I just tried the latest AOK release on TestFlight and it doesn't appear to have my PR. I think you merged the `master` branch in my fork which is not [the PR](https://github.com/zhaofengli/ish/tree/fstat-empty-path).

emkey1:
> @emkey1 I just tried the latest AOK release on TestFlight and it doesn't appear to have my PR. I think you merged the `master` branch in my fork which is not [the PR](https://github.com/zhaofengli/ish/tree/fstat-empty-path).

Thank you for the heads up and sorry about that.  I’ll take a look in a bit.

Nebucatnetzer:
```
/nix/store/acghbpn3aaj2q64mz3ljipsgf9d9qxlp-nix-2.11.1/bin/nix-store: error while loading shared libraries: libsodium.so.23: cannot stat shared object: No such file or directory
/tmp/nix-binary-tarball-unpack.XXXXaGPNbE/unpack/nix-2.11.1-i686-linux/install: unable to register valid paths
```

When I try to install it, I get the message above.
I installed `libsodium` and `libsodium-dev` but it still doesn’t work.

yassernasc:
nix on iSH? 👀

fs/stat.c
```diff
@@ -49,17 +69,18 @@ static struct fd *at_fd(fd_t f) {
     return f_get(f);
 }
 
-static dword_t sys_stat_path(fd_t at_f, addr_t path_addr, addr_t statbuf_addr, bool follow_links) {
+// The `flags` parameter accepts AT_ flags
+static dword_t sys_stat_path(fd_t at_f, addr_t path_addr, addr_t statbuf_addr, int flags) {
```
tbodt, line 73: This function now has the exact signature of sys_fstatat64 so maybe rename to sys_fstatat64

fs/stat.c
```diff
@@ -30,11 +30,31 @@ struct newstat64 stat_convert_newstat64(struct statbuf stat) {
...
     char path[MAX_PATH];
-    int err = path_normalize(at, path_raw, path, follow_links ? N_SYMLINK_FOLLOW : N_SYMLINK_NOFOLLOW);
-    if (err < 0)
-        return err;
+    if (empty_path && (strcmp(path_raw, "") == 0)) {
+        if (at == AT_PWD) {
+            at = current->fs->pwd;
+        }
+
+        err = generic_getpath(at, path);
```
tbodt, line 45: I suspect this should be calling fstat instead

tbodt (changes requested):

ghost (approved):

emkey1:
Checked this one against current master (7864dd60), and I think it can be closed as obsolete — reporting the measurements in case they save someone the same look.

**The functionality is already present on master.** `fs/stat.c:109` now has:

```c
if ((flags & AT_EMPTY_PATH_) && strcmp(path, "") == 0) {
```

so `generic_statat` handles the empty-path case, and `AT_EMPTY_PATH_` is defined in `kernel/fs.h`. It appears to have arrived by another route in the four years this has been open.

Verified with one static i386 binary run on real Linux and on master:

| | `fstatat(fd, "", AT_EMPTY_PATH)` | `fstatat(dirfd, "", AT_EMPTY_PATH)` |
|---|---|---|
| real Linux | `0`, size=5 | `0`, is a dir |
| upstream master | `0`, size=5 | `0`, is a dir |

Master matches Linux on both.

**The patch also no longer applies:**

```
error: patch failed: kernel/fs.h:44
error: kernel/fs.h: patch does not apply
```

So it would need a rebase to land, and there does not appear to be anything left for it to fix. @zhaofengli — thanks for it regardless; it was a real gap when you filed it.

saagarjha:
@emkey1 please do not respond to contributors as a maintainer

emkey1:
I'm confused.  Are you not familiar with how GitHub works?

saagarjha:
Yes? This is not a new policy by any means. You have been around for long enough to know that we do not really want backseat maintainers; see for example https://github.com/ish-app/ish/issues/1347#issuecomment-861384135. You are welcome to post in issues and PRs but providing nontechnical/procedural feedback to unrelated contributors should always be done in a way that does not leave confusion as to who is speaking. If this still does not make sense, I recommend you reach out and discuss with us so we can come to alignment :)

PR #1827, opened 2022-06-15
Alpine v3.14 -> v3.16

L1ghtmann:
I believe I got all of the relevant lines, but apologies in advance if I missed anything.


tbodt:
Going to close, in part because the changes for an upgrade are more involved than this, but also because an upgrade is risky and i don't want to do it yet.

Issue #1848, opened 2022-07-28
Ish_log


Beg1nner12:

Mnpn:
Hi again, it looks like you've forgotten to add any details which describe your issue! We can reopen it if that changes.

PR #1849, opened 2022-07-28
Update iSH.xcconfig

Beg1nner12:


tbodt:
why?

Rexogamer:
They've also created #1848, an empty issue

PR #1861, opened 2022-08-06
Speedup proc_print_escaped()

jason-conway:
Also check that `str` is nonnull before dereference


saagarjha (approved):
LGTM, aside from the nit that we don't really need a null check here. The caller should do that validation themselves

jason-conway:
Figured f286240 wasn't worth a separate PR; lmk if I should revert it

saagarjha (changes requested):
These code paths are not really performance-critical, so we'd like the balance for this code to lean towards readability and maintainability rather than maximum speed. In this case it would be appropriate to rely on whatever the optimizing compiler is willing to emit unless there's a good reason otherwise.

saagarjha (approved):

tbodt:
Please rebase on master to fix CI

PR #1863, opened 2022-08-08
Add sys_membarrier

jason-conway:
Handles `MEMBARRIER_CMD_QUERY` and throws everything else at a thread fence... low risk, low reward.

Also stubs sys_getcpu() to allow stress-ng to run


me355:
Shah

me355:
Noneo8

saagarjha:
membarrier is intended to be a memory barrier across all threads of the process, not just the task that called it. I haven't looked at what Linux is doing here but I suspect it involves using signals to support correctly on macOS, or parking the thread.

jason-conway:
may revisit this in the future, but closing this out for now

PR #1871, opened 2022-08-11
Add more mmx/sse2 instructions

jason-conway:
Adds [pmullw](https://www.felixcloutier.com/x86/pmullw) and [pmulhw](https://www.felixcloutier.com/x86/pmulhw) for both 64-bit and 128-bit operands.


tbodt:
Please enable the tests for these in qemu-test.c. You'll also need to run the test binary on real hardware to update the golden file.

tbodt:
Thanks!

PR #1880, opened 2022-08-23
Fix GDB

saagarjha:
Newer versions of GDB read from /proc/pid/task.

### DIFF
--- a/kernel/ptrace.c
+++ b/kernel/ptrace.c
@@ -119,12 +119,7 @@ dword_t sys_ptrace(dword_t request, dword_t pid, addr_t addr, dword_t data) {
             struct task *child = find_child(pid);
             if (!child) return _EPERM;
 
-            read_wrlock(&child->mem->lock);
-            char *ptr = mem_ptr(child->mem, addr, MEM_WRITE_PTRACE);
-            read_wrunlock(&child->mem->lock);
-            if (ptr) {
-                memcpy(ptr, &data, sizeof(data));
-            } else {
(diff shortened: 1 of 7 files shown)


kernel/ptrace.c
```diff
@@ -119,12 +119,7 @@ dword_t sys_ptrace(dword_t request, dword_t pid, addr_t addr, dword_t data) {
             struct task *child = find_child(pid);
             if (!child) return _EPERM;
 
-            read_wrlock(&child->mem->lock);
```
tbodt, line 122: why change this? was it broken?

tbodt: oh i forgot about the /proc/pid/mem, never mind

kernel/user.c
```diff
@@ -61,7 +68,7 @@ int user_read_string(addr_t addr, char *buf, size_t max) {
     read_wrlock(&current->mem->lock);
     size_t i = 0;
     while (i < max) {
-        if (__user_read_task(current, addr + i, &buf[i], sizeof(buf[i]))) {
+        if (__user_read_task(current, addr + i, &buf[i], sizeof(buf[i])), false) {
```
tbodt, line 71: uh does this compile? no parameter was added to __user_read_task

saagarjha: You're reading it wrong, this is using the comma operator

tbodt: this seems like a bad idea, if you're going to ignore the return value, why put the call in the if at all? very easy to read wrong as you've seen

saagarjha: I think you're still reading it wrong

PR #1881, opened 2022-08-23
stuff

saagarjha:


app/TerminalViewController.m
```diff
@@ -325,16 +325,23 @@ - (void)keyboardDidSomething:(NSNotification *)notification {
...
         return;
     NSLog(@"%@ %@", notification.name, [NSValue valueWithCGRect:keyboardFrame]);
     self.hasExternalKeyboard = keyboardFrame.size.height < 100;
-    CGFloat pad = UIScreen.mainScreen.bounds.size.height - keyboardFrame.origin.y;
-    if (pad != keyboardFrame.size.height) {
-        pad = 0; // keyboard is not right at the bottom of the screen, must be floating or something
-    }
-    if (pad == 0) {
-        pad = self.view.safeAreaInsets.bottom;
+    CGFloat pad = self.view.bounds.size.height - keyboardFrame.origin.y;
```
tbodt, line 333: Honestly the only way I can really review this is ask what ios versions you tested it on

app/AboutViewController.m
```diff
@@ -58,7 +58,7 @@ - (void)viewDidLoad {
             [self _updateUI];
         });
     }];
-    [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(_updateUI) name:FsUpdatedNotification object:nil];
+    [NSNotificationCenter.defaultCenter addObserver:self selector:@selector(_updateUI:) name:FsUpdatedNotification object:nil];
```
tbodt, line 61: How important is this?

app/TerminalViewController.m
saagarjha: iOS 12-16, and I got @Mnpn to do iOS 11

app/AboutViewController.m
saagarjha: In theory the selector takes a parameter that you are just ignoring completely from the signature, which is bad. In practice it's been working, of course.

PR #1893, opened 2022-09-02
Miscellaneous fixes

saagarjha:
These fell out of additional manual testing and user feedback for some recent changes.

### DIFF
--- a/app/RootsTableViewController.m
+++ b/app/RootsTableViewController.m
@@ -74,7 +74,7 @@
 
 - (void)documentPicker:(UIDocumentPickerViewController *)controller didPickDocumentsAtURLs:(NSArray<NSURL *> *)urls {
     NSAssert(urls.count == 1, @"somehow picked multiple documents");
-    NSURL *url = urls[0];
+    NSURL *url = urls.firstObject;
     NSString *fileName = url.lastPathComponent.stringByDeletingPathExtension;
     if ([fileName hasSuffix:@".tar"])
         fileName = fileName.stringByDeletingPathExtension;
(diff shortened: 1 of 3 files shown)


app/RootsTableViewController.m
```diff
@@ -90,7 +90,11 @@ - (void)documentPicker:(UIDocumentPickerViewController *)controller didPickDocum
 
     dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^{
         NSError *error;
-        BOOL success = [Roots.instance importRootFromArchive:urls[0] name:name error:&error progressReporter:progressVC];
+        BOOL success = NO;
+        if ([url startAccessingSecurityScopedResource]) {
```
tbodt, line 94: error is used uninitialized if this is false

saagarjha: It would be nil which should be fine in this case?

tbodt: it wouldn't, but realized that success is false so this is fine. Still, should we show something to the user if this fails? Not sure what failure here actually means though

saagarjha: Oh I just spotted the `error != nil` check, we can delete that

tbodt: Not sure if importRootFromArchive can maybe return false with error null?

saagarjha: In any case when something fails we should tell the user that the import didn't succeed, even if we don't have any additional error information

tbodt: Makes sense, but we should do that without possibly reading uninitialized data

saagarjha: I'm not seeing the uninitialized read

tbodt: startAccessingSecurityScopedResource returns no, then success is no, but error was never set and will be read in the if below, and then maybe by presentError

saagarjha: Right, it will be read as nil which is fine here

tbodt: Today I learned objective-c auto initializes object pointer variables to nil https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithObjects/WorkingwithObjects.html#//apple_ref/doc/uid/TP40011210-CH4-SW22

PR #1898, opened 2022-09-05
Add missing move instructions

jason-conway:
movntq and movntdq are supposed to use non-temporal hints... this does not do that


saagarjha:
Not the effects of the cache should be generally ok, FWIW

filiphsps:
Should probably squash the fix commit and rebase instead of merging master to prevent merge commits.

Other than that this looks great 👍

jason-conway:
Closing in lieu of #1926

PR #1903, opened 2022-09-07
Fix/webview terminal size spacing

davemackintosh:
This PR allows the web view to be the full width of the screen. This is only really noticeable when an app has a different background colour to the ish theme but it's annoying either way.

### DIFF
--- a/app/TerminalView.m
+++ b/app/TerminalView.m
@@ -118,6 +118,7 @@ static NSString *const HANDLERS[] = {@"syncFocus", @"focus", @"newScrollHeight",
     }
 
     WKWebView *webView = _terminal.webView;
+    CGSize screen = UIScreen.mainScreen.bounds.size;
     _terminal.enableVoiceOverAnnounce = YES;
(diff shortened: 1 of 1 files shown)


app/TerminalView.m
```diff
@@ -127,7 +128,7 @@ - (void)installTerminalView {
     for (int i = 0; i < sizeof(HANDLERS)/sizeof(HANDLERS[0]); i++) {
         [webView.configuration.userContentController addScriptMessageHandler:handler name:HANDLERS[i]];
     }
-    webView.frame = self.bounds;
+    webView.frame = CGRectMake(0, 0, screen.width + 13, self.bounds.size.height + 12);
```
davemackintosh, line 131: I'm not sure why the reported screen size is smaller than it is. I double checked the safe areas and bounds of the superviews but no idea.

davemackintosh:
This PR has introduced a few bugs so I'm closing it and will revisit in the near future with some other theme/layout based changes I'm working on.

PR #1917, opened 2022-09-26
Implement FCMOVcc, cmpss, pminsw, and pmaxsw

jason-conway:
Breaking this out from #1898 since their tests are covered by `qemu-test.c` as is


Aungkhant4 (commented):
K

PR #1933, opened 2022-10-24
Replace chatty syscall stubs with silent variant

saagarjha:
These syscalls have fallbacks in musl which we do implement but they are only invoked after they return ENOSYS, which fills dmesg with lots of spam.

### DIFF
--- a/kernel/calls.c
+++ b/kernel/calls.c
@@ -6,10 +6,16 @@
 #include "kernel/signal.h"
 #include "kernel/task.h"
 
-dword_t syscall_stub() {
+dword_t syscall_stub(void) {
     return _ENOSYS;
 }
-dword_t syscall_success_stub() {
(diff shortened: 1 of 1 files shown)


kernel/calls.c
```diff
@@ -6,10 +6,13 @@
 #include "kernel/signal.h"
 #include "kernel/task.h"
 
-dword_t syscall_stub() {
+dword_t syscall_stub(void) {
     return _ENOSYS;
 }
-dword_t syscall_success_stub() {
+dword_t syscall_silent_stub(void) {
```
tbodt, line 12: Should have a comment explaining how one of these is silent and one isn't. Or just move the stub syscall printk into the actual syscall stub function tbh

tbodt (approved):

PR #1940, opened 2022-10-30
added nullptr check (previously segfaulted)

KhankNyo:
Running ./ish with no command line argument resulted in a segmentation fault.

On line 106 in xX_main_Xx.h, the first argument of do_execve() received a nullptr, but no adequate checks were provided to handle nullptr. The nullptr is then passed into strcmp(), which resulted in undefined behaviors (segfaulted on my system).

So a check for (path == NULL) was provided in path_normalize(), so _ENOENT can be returned instead of undefined behavior from passing NULL into strcmp().

### DIFF
--- a/fs/path.c
+++ b/fs/path.c
@@ -120,7 +120,7 @@ static int __path_normalize(const char *at_path, const char *path, char *out, in
 
 int path_normalize(struct fd *at, const char *path, char *out, int flags) {
     assert(at != NULL);
-    if ((path == NULL) || (strcmp(path, "") == 0))
+    if ((strcmp(path, "") == 0))
         return _ENOENT;
 
     // start with root or cwd, depending on whether it starts with a slash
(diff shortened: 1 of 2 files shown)


saagarjha (commented):
I feel like the more appropriate fix here would be to add a check in `xX_main_Xx`, since it's responsible for passing a valid path into `do_execve`.

KhankNyo:
Thank you for the suggestion! I'll add a check for a null path right before the call to do_execve() instead of inside path_normalize().

fs/path.c
```diff
@@ -120,7 +120,7 @@ static int __path_normalize(const char *at_path, const char *path, char *out, in
 
 int path_normalize(struct fd *at, const char *path, char *out, int flags) {
     assert(at != NULL);
-    if (strcmp(path, "") == 0)
+    if ((strcmp(path, "") == 0))
```
tbodt, line 123: Why?

KhankNyo: Oops I forgot to remove the extra brackets, fixed it!

tbodt:
Thanjsk

PR #1958, opened 2022-11-12
meson: fix warnings

jason-conway:
Fixes the following:
```
WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.
         See also: https://github.com/mesonbuild/meson/issues/9300
```
and
```
WARNING: find_library('libarchive') starting in "lib" only works by accident and is not portable
```


tbodt:
Seems to break macos CI for some reason? Please fix

tbodt:
CI fixed itself I guess?

PR #1970, opened 2022-11-27
Revert "Return the number of bytes written from sched_getaffinity"

timwa0669:
https://github.com/ish-app/ish/issues/1947
This reverts commit 6da3e79fc39bb10902b364cf96207ef933802479.

I can't do anything but revert the commit caused this issue.
Maybe you can find a better solution:)


tbodt:
The commit in question fixed a bug, and the new behavior is correct, so I don't want to revert it. I suspect it revealed issues because go tried to use more concurrency. Try setting GOMAXPROCS=1?

timwa0669:
> The commit in question fixed a bug, and the new behavior is correct, so I don't want to revert it. I suspect it revealed issues because go tried to use more concurrency. Try setting GOMAXPROCS=1?

This also works with no error reported.

PR #2017, opened 2023-01-15
Patch out web view focus stealing first responder

### LINKED ISSUE #1937: Keyboard disappears on iOS 15.7
I am a blind user using voiceover screen reader.  After updating my device to iOs 15.7, when using iSH, I very often hide the keyboard as if I pressed the "Hide Keyboard" button, but I did not.  Because of this, it became simply impossible to use the program, since the keyboard disappears almost instantly after it was displayed again.  With a bluetooth keyboard, this problem unfortunately persists.

saagarjha:
This is a workaround for https://bugs.webkit.org/show_bug.cgi?id=249976, which should fix #1937.

There's a couple of things going on here that are probably worth noting:

* We rely on some things from the macOS SDK that happens to not be copied over the the iPhone one.
	* These are part of the platform ABI. In once case we use mig but could drop it in the future and use the generated files directly, if it becomes necessary. In another case use use a function from a header that Apple seems to have forgotten to include in the SDK. We can just ask for them to include it, and even if it goes away it isn't critical to anything.
* We need to parse the shared cache for symbols.
	* For space reasons Apple removes symbols from the address space that aren't exported. We need to read them out again from the file on disk. This file has a format that is subject to change. So far it has never changed in this way (I checked a decade). If it does this would break Apple's tooling, so there's a magic header that they check against to see if they know how to handle the file. We can check against this too.
* We need to set hardware breakpoints.
	* This is all API from the iPhone SDK, and it's used by the open-source LLDB. It's unusual to use it from current process, but ultimately all the function calls take a task and nothing is stopping us from offering up our own.
* We need to install a Mach exception handler.
	* This is API and pretty standard for crash reporting code, so it shouldn't break ever. One thing that has changed recently is that the guarded task ports mitigation makes it difficult to forward exceptions that we don't handle ourselves, which is useful when something else installs a handler at the thread/task level such as a debugger. If you have VoiceOver enabled and need the debugger to work it's not all that hard to disable the exception server. (I could also add code to do this automatically, but let's do that if it becomes a problem.)
* This entire thing is insane.
	* Yes but I've been unable to find another solution. Bugs have been filed and we'll continue to engage with Apple to see if we can improve our options. The code we are patching is not really something we should be touching, but I've put significant effort into finding the most specific API we can hook, making sure we immediately bail if anything seems to be going wrong, and only activate this code path on devices running new versions of iOS with VoiceOver enabled. Since those users already cannot really use iSH, this should be an improvement for them.

### DIFF
--- a/app/hook.c
+++ b/app/hook.c
@@ -1,0 +1,274 @@
+//
+//  hook.c
+//  iSH
+//
+//  Created by Saagar Jha on 12/29/22.
+//
+
+#include "hook.h"
+#include "mach_excServer.h"
+#include <CoreFoundation/CoreFoundation.h>
+#include <dlfcn.h>
+#include <mach-o/dyld_images.h>
+#include <mach-o/nlist.h>
+#include <mach/mach.h>
+#include <pthread.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/sysctl.h>
+
+#if __arm64__
+
+kern_return_t catch_mach_exception_raise(
+    mach_port_t exception_port,
+    mach_port_t thread,
+    mach_port_t task,
+    exception_type_t exception,
+    mach_exception_data_t code,
+    mach_msg_type_number_t codeCnt) {
+    abort();
+}
+
+kern_return_t catch_mach_exception_raise_state_identity(
+    mach_port_t exception_port,
+    mach_port_t thread,
+    mach_port_t task,
+    exception_type_t exception,
+    mach_exception_data_t code,
+    mach_msg_type_number_t codeCnt,
+    int *flavor,
+    thread_state_t old_state,
+    mach_msg_type_number_t old_stateCnt,
+    thread_state_t new_state,
+    mach_msg_type_number_t *new_stateCnt) {
+    abort();
+}
+
+static bool initialized;
+
+struct hook {
+    uintptr_t old;
+    uintptr_t new;
+};
+static struct hook hooks[16];
+static int active_hooks;
+static int breakpoints;
+
+mach_port_t server;
+
+kern_return_t catch_mach_exception_raise_state(
+    mach_port_t exception_port,
+    exception_type_t exception,
+    const mach_exception_data_t code,
+    mach_msg_type_number_t codeCnt,
+    int *flavor,
+    const thread_state_t old_state,
+    mach_msg_type_number_t old_stateCnt,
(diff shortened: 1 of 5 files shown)


app/hook.c
```diff
@@ -0,0 +1,274 @@
...
+
+void *find_symbol(void *base, char *symbol) {
+    if (!proc_regionfilename) {
+        proc_regionfilename = dlsym(dlopen(NULL, RTLD_LAZY), "proc_regionfilename");
+    }
+
+#define CHECK(x)         \
+    do {                 \
+        if (!(x)) {      \
+            return NULL; \
```
tbodt, line 147: Maybe we should log?

app/hook.c
```diff
@@ -0,0 +1,274 @@
...
+    mach_msg_server(mach_exc_server, sizeof(union __RequestUnion__catch_mach_exc_subsystem), server, MACH_MSG_OPTION_NONE);
+    abort();
+}
+
+static bool initialize_if_needed() {
+    if (initialized) {
+        return true;
+    }
+
+#define CHECK(x)          \
```
tbodt, line 94: Weird to see the same macro defined, undefined, then redefined

tbodt (approved):
I hate this

app/hook.c
tbodt: Never mind I just realized the return type is different

jacy7282898:
元の実装、おそらくトランポリンを保存していただけますか

saagarjha:
In theory it's probably possible? Would be a bit of work to implement correctly though. I think you'd probably have to disable the hardware breakpoint, enable single-stepping, and then reenable it? A simple trampoline seems like it wouldn't be enough.

jacy7282898:
>おそらく、ハードウェア ブレークポイントを無効にし、シングルステップを有効にしてから、再度有効にする必要があるでしょうか。シンプルなトランポリンでは物足りないようです。

この実装を非公開で追加するための寄付を受け付けていただけませんか 自由時間に

saagarjha:
Can you tell me more about what you’re trying to do? My email is in my profile if you would like to share this privately.

jacy7282898:
> Can you tell me more about what you’re trying to do? My email is in my profile if you would like to share this privately.

i threw an email, thanks

jacy7282898:
I have sent an email with my username in the title, thank you

PR #2036, opened 2023-02-07
Some additional MMX instructions

jason-conway:


tests/e2e/qemu/qemu-test.c
```diff
@@ -2306,6 +2306,22 @@ static uint64_t __attribute__((aligned(16))) test_values[4][2] = {
     }\
 }
 
+#define SHIFT_OP_MM(op,ib)\
```
tbodt, line 2309: Could this be added to SHIFT_OP instead of being its own macro?

jason-conway: Absolutely, but we'll need to add the remaining encodings for those instructions first. I'll go ahead and do that here now 👍

PR #2109, opened 2023-05-06
Encode exceptions into the crash log

saagarjha:
Apple's crash reports don't contain exception name/reason information. We can get this back in a roundabout way by constructing a fake backtrace and encoding the exception information into that.

### DIFF
--- a/app/ExceptionExfiltrator.m
+++ b/app/ExceptionExfiltrator.m
@@ -1,0 +1,290 @@
+//
+//  ExceptionExfiltrator.m
+//  libiSHApp
+//
+//  Created by Saagar Jha on 5/5/23.
+//
+
+#import "ExceptionExfiltrator.h"
+#import <Foundation/Foundation.h>
+#import <execinfo.h>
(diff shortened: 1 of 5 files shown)


app/ExceptionExfiltrator.m
```diff
@@ -0,0 +1,290 @@
...
+#import <execinfo.h>
+#import <pthread.h>
+
+#define f(name, character)         \
+	__asm__("\" " name "\": nop"); \
+	void ish_exception_exfiltrate_##character(void) __asm__(" " name)
+
+f("unprintable", unprintable);
+
+f(" ", space);
```
tbodt, line 19: Can we use X macros to fix the duplication

tbodt (commented):
Call this from ios_handle_die too

app/ExceptionExfiltrator.m
saagarjha: Probably not

seabass5vg8 (commented):
_null_

PR #2129, opened 2023-06-05
Quickfix x86 builds

0x41c:
A small patch for the time being until the build environment gets rewritten. I attempted to have `EXCLUDED_ARCHS` set from the xcconfig, but it just wouldn't take and I'd still get issues when it'd try cross compiling the arm64 build. I've left a comment about that unless there's another way of getting the following line to work:

```xcconfig
EXCLUDED_ARCHS[sdk=macosx*][arch=x86_64] = arm64
```

This is all under the assumption that universal builds just won't make the cut for iSH.


62f:
> 2129 @randomnessExistance No, more like so:

/pull/2129

Robertearl5770:
This 2 file changed was that on the ish shell ?

TichThong (approved):

AimSD23nSiR (commented):
Random

PR #2352, opened 2024-02-24
fix ish-app/ish#2349 stop EACCES truncate with open

kuflierl:
Relocate the permission check for the `generic_openat` function to before opening the back-end.
This mitigates the case where open is called on the file back-end even if the user is not allowed to.
Opening the back-end prematurely can cause truncation of the file in question by the error handler.

This seemed like an easy issue to fix so i tried fixing it.
This is my first PR in this Repository, If i made any mistakes let me know!

### DIFF
--- a/fs/generic.c
+++ b/fs/generic.c
@@ -43,8 +43,10 @@ struct fd *generic_openat(struct fd *at, const char *path_raw, int flags, int mo
         return ERR_PTR(err);
     struct mount *mount = find_mount_and_trim_path(path);
     
+    lock(&inodes_lock); // TODO: don't do this
     // get file stats before opening file to facilitate permission checking before the backend opens the file
     // opening a file with truncate and then closing it will cause the file contents to be cleared
+    // this might in theory cause a Time-of-Check to Time-of-Use privilage escalation by replacing the file with a symlink.
     struct statbuf stat;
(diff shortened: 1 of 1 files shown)


tbodt:
Looks like this makes tests fail

kuflierl:
That's indeed worrying, I will look into it when I have time

kuflierl:
> Looks like this makes tests fail

This time it should work @tbodt  I ran the tests.

I had to engage the inode lock before the stat

kuflierl:
We may need no increase the timeout delay to reduce false positives

kuflierl:
Actually there might be a better way that doesn't cripple the speed. but it might include a slight rework of the fd closing function. @tbodt Would you prefer that?

tbodt:
What happened with the speed?

If more things need to be refactored, more things need to be refactored :D This isn't a bad thing in itself.

kuflierl:
> What happened with the speed?
> 
> If more things need to be refactored, more things need to be refactored :D This isn't a bad thing in itself.

It seems fstat is that much faster than stat. I expected a bit of performance loss due to extra path finding overhead but not this much. Instead of checking before opening the file it might be a better idea to just write a better error handling function to compensate.

Jk2987 (approved):

brahunt23 (approved):

PR #2357, opened 2024-03-06
Update and rename SECURITY.md to SECURITY/HACKERS/delete

XxSonicGhostxX:
@security.github


saagarjha:
Hi! We appreciate your interest in the project’s security model but unfortunately your contribution does not rise to the level of something we would be looking to incorporate into our policy. iSH is explicitly not designed for production use, so it has a very different perspective from most other open source software. Our security documentation is intended to communicate this, and we welcome feedback if any part of it seems unclear or incorrect. If you nonetheless believe that your device has been compromised as a result of our security posture please provide as much detail as possible and our team will be happy to look into it for you.

Issue #711, opened 2020-05-09
Emacs broke


tbodt:
https://discord.com/channels/508839261924229141/510973900209913857/708708134398459905

> Help me! I installed ish on my ipad air3, after ran freshly installed emacs I got this error “Failed select: Invalid argument”. I have tried  ‘emacs —no—init-file’ , ‘apk upgrade & apk update’, reinstall emacs, reinstall ish, but with no success. After some googling I found this [https://github.com/termux/termux-packages/issues/952] and this [https://github.com/termux/termux-packages/pull/956]. Somebody encountered a bug in emacs under termux, and it looks like the same bug which ish happens to have?

glyph:
I saw this same issue on “1.0 (67)” via testflight. Emacs was the first thing I tried to use after installing, and this was a bit sad. So close!!

tbodt:
The good news is this is fixed if you build from source, and is also a regression in just the latest version, so you can try installing the one before that.

glyph:
Wow, thanks for the instant feedback! Now the only problem is that control-space doesn’t seem to be accessible on the iPad keyboard, and that’s a pretty important key binding in Emacs :).

tbodt:
That would be good to file a new issue for, in the meantime you can probably fake it with ctrl-2

PR #2359, opened 2024-03-08
Fix bad timeout values in `pselect()` (which broke mosh)

cgull:
`mosh` almost works, but it has weird unpredictable 1-30 second lag on text input that is quite incredibly annoying.

I ran this down to a fairly simple oversight:  `pselect()` takes a `timespec` for the timeout value, but that was getting passed unchanged to the iSH `select()` implementation which takes a `timeval`.

It looks like a fix was made for #711 earlier that sort of worked; this probably fixes the real underlying problem with Emacs.  That change clamped the excessively big nanosecond values to be +1 second, but Mosh was still badly affected.

This will probably clear up other mysterious lags in other applications.

There's a couple of other minor fixes I needed to get the CLI build working and logging, too.

### DIFF
--- a/kernel/poll.c
+++ b/kernel/poll.c
@@ -33,7 +33,8 @@ static int select_event_callback(void *context, int types, union poll_fd_info in
         return 0;
     return 1;
 }
-dword_t sys_select(fd_t nfds, addr_t readfds_addr, addr_t writefds_addr, addr_t exceptfds_addr, addr_t timeout_addr) {
+
+static dword_t select_common(fd_t nfds, addr_t readfds_addr, addr_t writefds_addr, addr_t exceptfds_addr, struct timespec *timeout_addr, const char *name) {
     size_t fdset_size = BITS_SIZE(nfds);
     char readfds[fdset_size];
     if (user_read_or_zero(readfds_addr, readfds, fdset_size))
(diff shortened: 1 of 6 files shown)


kernel/poll.c
```diff
@@ -46,14 +46,24 @@ dword_t sys_select(fd_t nfds, addr_t readfds_addr, addr_t writefds_addr, addr_t
         return _EFAULT;
 
     struct timespec timeout_ts = {};
-    if (timeout_addr != 0) {
+    if (timeout_addr != 0 && timeout_ts_addr != 0) {
+        assert(!"select_internal BUG");
```
tbodt, line 50: We can prevent this error at compile time instead of runtime. Let's move the code to read and convert timeval to the caller, removing the need for two mutually exclusive arguments. There are also [utility functions](https://github.com/ish-app/ish/blob/33eaf3bd2c0402309310b6b518edda2300c9df61/kernel/fs.c#L735C1-L747C2) already written for similar purposes, we should move them somewhere like util/timer.h and use them.

tools/meson.build
```diff
@@ -25,11 +25,14 @@ libarchive = dependency('libarchive', required: false)
 if not libarchive.found()
     # homebrew
     libarchive_lib = cc.find_library('archive', dirs: ['/usr/local/opt/libarchive/lib'], required: false)
-    if not libarchive_lib.found()
```
tbodt, line 28: The problem is just include_directories, right? Instead of this, let's make a variable similar to libarchive_lib for the different include directories. Something like

```
libarchive_lib = ...
libarchive_includes = ['/usr/local/opt/libarchive/include']
if not libarchive_lib.found()
	libarchive_lib = ...
	libarchive_includes = ['/opt/homebrew/opt/libarchive/lib']
endif
```

tbodt (changes requested):
Good find and thanks for the fix! I have a number of stylistic comments.

kernel/poll.c
cgull: Done.

tools/meson.build
cgull: Not quite, I think?  Depends on what you mean.  The problem was that whether your Mac has Homebrew in `/usr/local/opt` (x86) or `/opt` (Apple Silicon), it was always choosing `/usr/local/opt/libarchive/include`.  That's obviously inconsistent on Apple Silicon, and it doesn't work on my Apple Silicon machine.  Maybe your machine has a compatibilty symlink for that and mine doesn't?

Also, the existing code allows builds (without fakefsify) when libarchive is not present, I think your proposed change would break that.

I left this change as-is for now but I'll revise if you still want.

tbodt:
Thanks!

PR #2376, opened 2024-04-24
Remove "engines" and drop interpreter code

saagarjha:
This hasn't worked for a very long time, and the "JIT" engine is basically all we test so we should just turn it on by default.


tbodt:
I'd like to keep the engine meson option, because we will inevitably add other interpreter implementations and want to switch at build time, but the rest is good.

PR #2386, opened 2024-05-11
Avoid crashes on /proc/ish from CLI build

cgull:
`find /proc` segfaults on CLI iSH.

The CLI iSH does not implement some callouts that iSH's procfs provides for the GUI, and the procfs does not check that these callouts actually exist.  This fixes that.

### DIFF
--- a/fs/proc/ish.c
+++ b/fs/proc/ish.c
@@ -55,6 +55,10 @@ static int proc_ish_show_colors(struct proc_entry *UNUSED(entry), struct proc_da
 }
 
 static int proc_ish_show_documents(struct proc_entry *UNUSED(entry), struct proc_data *buf) {
+    if (get_documents_directory == NULL) {
+        proc_printf(buf, "\n");
+        return 0;
+    }
     char *directory = get_documents_directory();
     proc_printf(buf, "%s\n", directory);
     free(directory);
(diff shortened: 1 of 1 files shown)


fs/proc/ish.c
```diff
@@ -55,6 +55,10 @@ static int proc_ish_show_colors(struct proc_entry *UNUSED(entry), struct proc_da
 }
 
 static int proc_ish_show_documents(struct proc_entry *UNUSED(entry), struct proc_data *buf) {
+    if (get_documents_directory == NULL) {
+	proc_printf(buf, "\n");
+	return 0;
```
tbodt, line 60: Please fix indentation (is this perhaps using tabs instead of spaces or something?

tbodt (commented):
Looks mostly good, other than indentation. Thanks!

fs/proc/ish.c
cgull: Yes, tabs.  You may see these in C code occasionally.  :)

You might want to add some style guidance to the docs, or enforce with `clang-format`, if it's important to you.

Also fixed in #2387

tbodt: The problem with it is that github renders tabs as four spaces, while your editor probably used eight... thus are the hazards of mixing indentation types.

PR #2387, opened 2024-05-11
Fix pty close behavior, signals at exit

cgull:
There's two big things in here.

When running `mosh --local 0`, I found that `mosh-server` would never terminate on iSH.  It was waiting for the master side of its pty to close.  Most applications that use ptys use other methods to detect child termination, but `mosh-server` expects that the pty master will act as a closed file after the last program on the pty slave exits-- any waiting reads should return immediately with 0 bytes, and further read calls should return EOF.  This is what pipes and sockets do, of course.  iSH master ptys stayed in the read call forever, even though the pty slave was gone.  This fixes that and ends up better separating generic tty code from pty slaves from pty masters too.

`dtach` behaves similarly and the fix helps it too.

After I fixed that, I found that CLI ish would exit with signal 9 when `mosh-server` exited.  This is because the README recommends to use `/bin/login` as the starting process.  Unfortunately, that means `/bin/login` runs as pid 1 (init) and is notified when orphaned processes terminate (which `mosh-server` is, because it double forks).  `/bin/login` has never been intended as a process group leader or init process, it expects to start and `wait()` for exactly 1 child, usually an interactive shell.

But running `ish -f alpine /bin/sh` didn't work correctly either, when the initial shell exited iSH would exit on signal 9. (Or something, this was a couple of months ago and I'm forgetting details.  Maybe it was still when `mosh-server` exited, or when a subshell exited.)  That turned out to be a nasty problem with `halt_system()`.  This code runs when pid 1 exits.  This code uses `pthread_kill()` to send a SIGKILL to threads running the emulated process group to kill off emulated processes in iSH.  Uhh... that doesn't work.  SIGKILL always nukes the entire process, not a single thread.

After I changed that code to do a more kernel-like thing of using `deliver_signal()` to send first SIGTERM, then SIGKILL, and then finally `pthread_kill()` with SIGTERM, iSH behaves a lot better about emulated process termination, and exits cleanly (rather than with SIGKILL) when pid 1 terminates.

I changed the README to tell people to use `/bin/sh` as the starting process for CLI ish, because it can work as init/pid 1 and ignores termination of orphaned processes.  iSH.app should probably change to do the same.

I've not been able to build iSH.app in XCode, so all of this is untested there, but I believe it will help there-- the app won't quit when the initial shell exits.

After fixing this and then finding/fixing the procfs problem in #2386, I ran into yet another bug in iSH, I forget what it was.  At that point I gave up chasing bugs in iSH.

### DIFF
--- a/fs/pty.c
+++ b/fs/pty.c
@@ -34,12 +34,30 @@ static int pty_master_init(struct tty *tty) {
     return 0;
 }
 
+
+static void pty_hangup(struct tty *tty) {
+    if (tty == NULL)
+        return;
+    lock(&tty->lock);
+    tty_hangup(tty);
+    unlock(&tty->lock);
+}
+
+static struct tty *pty_hangup_other(struct tty *tty) {
+    struct tty *other = tty->pty.other;
+    if (other == NULL)
+        return NULL;
+    pty_hangup(other);
+    return other;
+}
+
+static void pty_slave_cleanup(struct tty *tty) {
+    pty_hangup_other(tty);
+}
+
 static void pty_master_cleanup(struct tty *tty) {
-    struct tty *slave = tty->pty.other;
+    struct tty *slave = pty_hangup_other(tty);
     slave->pty.other = NULL;
-    lock(&slave->lock);
-    tty_hangup(slave);
-    unlock(&slave->lock);
     tty_release(slave);
 }
 
@@ -48,6 +66,16 @@ static int pty_slave_open(struct tty *tty) {
         return _EIO;
     if (tty->pty.locked)
         return _EIO;
+    return 0;
+}
+
+static int pty_slave_close(struct tty *tty) {
+    // If userland's reference count on the pty slave will go to 0,
+    // hang up the pty master.  But the session leader may have a
+    // reference, and the pty master always has a reference.
+    if (tty->refcount - 1 == (tty->session ? 2 : 1)) {
+        pty_hangup_other(tty);
+    }
     return 0;
 }
 
@@ -94,7 +122,9 @@ DEFINE_TTY_DRIVER(pty_master, &pty_master_ops, TTY_PSEUDO_MASTER_MAJOR, MAX_PTYS
 const struct tty_driver_ops pty_slave_ops = {
     .init = pty_return_eio,
     .open = pty_slave_open,
+    .close = pty_slave_close,
     .write = pty_write,
+    .cleanup = pty_slave_cleanup,
 };
 DEFINE_TTY_DRIVER(pty_slave, &pty_slave_ops, TTY_PSEUDO_SLAVE_MAJOR, MAX_PTYS);
 
--- a/fs/tty.c
+++ b/fs/tty.c
@@ -672,7 +672,7 @@ static int tiocgpgrp(struct tty *tty, pid_t_ *fg_group) {
         lock(&slave->lock);
     }
 
-    if (tty == slave && !tty_is_current(slave) || slave->fg_group == 0) {
+    if (tty == slave && (!tty_is_current(slave) || slave->fg_group == 0)) {
         err = _ENOTTY;
         goto error_no_ctrl_tty;
     }
--- a/fs/tty.c
+++ b/fs/tty.c
@@ -110,16 +110,7 @@ void tty_release(struct tty *tty) {
         cond_destroy(&tty->produced);
         free(tty);
     } else {
-        // bit of a hack
-        struct tty *master = NULL;
-        if (tty->driver == &pty_slave && tty->refcount == 1)
-            master = tty->pty.other;
         unlock(&tty->lock);
-        if (master != NULL) {
-            lock(&master->lock);
-            tty_poll_wakeup(master, POLL_READ | POLL_HUP);
-            unlock(&master->lock);
-        }
     }
 }
 
@@ -207,11 +198,14 @@ static int tty_device_open(int major, int minor, struct fd *fd) {
 
 static int tty_close(struct fd *fd) {
     if (fd->tty != NULL) {
-        lock(&fd->tty->fds_lock);
+        struct tty *tty = fd->tty;
+        lock(&tty->fds_lock);
         list_remove_safe(&fd->tty_other_fds);
-        unlock(&fd->tty->fds_lock);
+        unlock(&tty->fds_lock);
         lock(&ttys_lock);
-        tty_release(fd->tty);
+        if (tty->driver->ops->close)
+            tty->driver->ops->close(tty);
+        tty_release(tty);
         unlock(&ttys_lock);
     }
     return 0;
@@ -414,7 +408,7 @@ static bool pty_is_half_closed_master(struct tty *tty) {
     struct tty *slave = tty->pty.other;
     // only time one tty lock is nested in another
     lock(&slave->lock);
-    bool half_closed = slave->ever_opened && slave->refcount == 1;
+    bool half_closed = slave->ever_opened && (slave->refcount == 1 || slave->hung_up);
     unlock(&slave->lock);
     return half_closed;
 }
@@ -450,7 +444,7 @@ static ssize_t tty_read(struct fd *fd, void *buf, size_t bufsize) {
     struct tty *tty = fd->tty;
     lock(&pids_lock);
     lock(&tty->lock);
-    if (tty->hung_up) {
+    if (tty->hung_up || pty_is_half_closed_master(tty)) {
         unlock(&pids_lock);
         goto error;
     }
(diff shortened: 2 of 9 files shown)


cgull:
Tabs removed.

kernel/log.c
```diff
@@ -154,6 +154,10 @@ static void log_line(const char *line) {
 static void log_line(const char *line) {
     os_log_fault(OS_LOG_DEFAULT, "%s", line);
 }
+#elif LOG_HANDLER_STDERR
```
tbodt, line 157: I'm unclear on the utility of this since the program in ish may also be writing to stderr, but if you found it useful it doesn't hurt to have.

fs/pty.c
```diff
@@ -51,6 +69,13 @@ static int pty_slave_open(struct tty *tty) {
     return 0;
 }
 
+static int pty_slave_close(struct tty *tty) {
+    if (tty->refcount - 1 == (tty->session ? 2 : 1)) {
```
tbodt, line 73: This looks like a really tricky bit. Could you add a comment?

fs/tty.c
```diff
@@ -110,16 +110,7 @@ void tty_release(struct tty *tty) {
         cond_destroy(&tty->produced);
         free(tty);
     } else {
-        // bit of a hack
```
tbodt, line 113: Bravo for cleaning this up

fs/tty.c
```diff
@@ -414,7 +408,7 @@ static bool pty_is_half_closed_master(struct tty *tty) {
     struct tty *slave = tty->pty.other;
     // only time one tty lock is nested in another
     lock(&slave->lock);
-    bool half_closed = slave->ever_opened && slave->refcount == 1;
+    bool half_closed = slave->ever_opened && (slave->refcount == 1 || slave->hung_up);
```
tbodt, line 411: I wonder if the refcount check is still necessary? Would the new pty_slave_close do that?

tbodt (commented):
Thanks for working on this!

fs/tty.c
cgull: This is needed when `pty_slave_cleanup()` runs, I think.  It calls `pty_hangup_other()`, which calls `tty_hangup()` on the master.  That calls `tty_input_wakeup()` on the master, which may result in that user program waking up in `tty_read()`, or calling `read()` again, before the pty slave device instance is fully cleaned up and the refcount decremented.  That's what I remember, anyway.

kernel/log.c
cgull: If iSH is running interactively (stdin and stdout are on a tty), then emulated programs are running with both stdout and stderr on the iSH tty driver.  (This will be true for iSH.app as well, but writing to stderr might be less useful there.)

I hadn't even thought of CLI iSH supporting non-interactive use but yes, it does support that, and logging to stderr would be inappropriate then.

fs/pty.c
cgull: Done.

tbodt (approved):

francoisvignon:
Hi.

It seem the new behavior suffer an issue:
once the client close his side of the pty, no new connection can be issued to the provider without closing and reopening the provider

An idea ?

____________________

Francois

tbodt:
@francoisvignon do you have an easy way to reproduce? @cgull could you check on this?

francoisvignon:
here is an easy way to reproduce.

using iSH 669  and :
- gcc-11.2.1_git20220219-r2 x86 {gcc} (GPL-2.0-or-later LGPL-2.1-or-later) [installed]
- libgcc-11.2.1_git20220219-r2 x86 {gcc} (GPL-2.0-or-later LGPL-2.1-or-later) [installed]
- libstdc++-11.2.1_git20220219-r2 x86 {gcc} (GPL-2.0-or-later LGPL-2.1-or-later) [installed]
- minicom-2.8-r0 x86 {minicom} (GPL-2.0-or-later) [installed]
- musl-dev-1.2.3-r3 x86 {musl} (MIT) [installed]
- musl-utils-1.2.3-r3 x86 {musl} (MIT BSD GPL2+) [installed]
- make-4.3-r0 x86 {make} (GPL-3.0-or-later) [installed]

source: main.c
~~~
#include <stdio.h>
#include <errno.h>
#include <fcntl.h>
#include <stdlib.h>
#include <unistd.h>
  
int main(int argv, char **argc)
{
	int fd;
	fd = posix_openpt(O_RDWR | O_NOCTTY);
	grantpt(fd);
	unlockpt(fd);
	printf("client name %s\n", ptsname(fd));

	for(;;)
	{
		char buf[256];
		int l;
		l = read(fd, buf, 256);
		printf("l=%d, errno=%d\n", l, errno);
		sleep(1); // to not have too many traces ;-)
	}
	return(0);
}
~~~

compiling by : 
~~~
make main
~~~
executing:
~~~
McFv:~/test# ./main 
client name /dev/pts/2
// from other iSH window, connect to /dev/pts/2 by : minicom -D /dev/pts/2
// successfull
// keying CR one and two times
l=1, errno=0
l=2, errno=0
// disconnecting minicom by ^A ^X
l=-1, errno=5
l=0, errno=5
l=0, errno=5
l=0, errno=5
// from other iSH window, trying to reconnect by : minicom -D /dev/pts/2
// no way ...
l=0, errno=5
l=0, errno=5
l=0, errno=5
l=0, errno=s
// killing main
^C
McFv:
// and restarting main
McFv:~/test# ./main
client name /dev/pts/2 
// from other iSH window, connect to /dev/pts/2 by : minicom -D /dev/pts/2
// successfull
// keying CR one time
1=1, errno=0
// disconnecting minicom by ^A ^X
l=-1, errno=5
l=0, errno=5
l=0, errno=5
l=0, errno=5
l=0, errno=5
...
~~~


the same test with iSH 614
~~~
McFv:~/test# ./main
client name /dev/pts/2
// connect to /dev/pts/2 by : minicom -D /dev/pts/2
// successfull
// keying CR one and two times
l=1, errno=0
l=2, errno=0
// disconnecting minicom by ^A ^X
// read not interrrupted by disconnection, so, as read is blocking ... no new traces
// from other iSH window, trying to reconnect by : minicom -D /dev/pts/2
// successfull
// keying CR one and two times
l=1, errno=0
// killing main
^C
// minicom show /dev/pts/2 is not longer available
McFv:
~~~

with 614, client disconnection can be detected by other mean (setting read not blocking and checking errno).

please note, the source is voluntary simple to be the shortest code showing the issue (as example, displaying errno if not error is voluntary to not have too many lines, not setting read as not blocking also, ... ).

@cgull : I'm available to do more testing. thanks for your worlk

tbodt:
Ah, one bug replaced by another... The behavior on linux is different from either of those.

```
$ ./main
client name /dev/pts/1
// connect, type stuff
l=1, errno=0
l=3, errno=0
l=1, errno=0
l=6, errno=0
l=6, errno=0
// disconnect, get error 5 repeating
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
// reconnect, errors stop (well it still says `errno=5` but since the actual return code was success ignore that)
l=1, errno=5
l=2, errno=5
// disconnect, errors are back
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
l=-1, errno=5
```

francoisvignon:
this test was done with what ?

____________________

Francois

tbodt:
Linux linux-luka 6.7.6-1-aarch64-ARCH #1 SMP PREEMPT_DYNAMIC Sat Feb 24 10:08:35 MST 2024 aarch64 GNU/Linux

francoisvignon:
ok, this is the "expected" behavior :-)

____________________

Francois

tbodt:
I'm trying to decide if I should revert this PR, if the new bug is worse than the old bug. Obviously ideally both bugs would be fixed, but I don't really have time to do that work.

francoisvignon:
I will check the old bug tomorrow: I have used the old pty implementation without major issue ... so, I will double check, and I tell you ...send from my mobile.____________________FrancoisLe 18 mai 2024 à 18:35, tbodt ***@***.***> a écrit :﻿
I'm trying to decide if I should revert this PR, if the new bug is worse than the old bug. Obviously ideally both bugs would be fixed, but I don't really have time to do that work.

francoisvignon:
Major ? annoying is a better word:- with the previous implementaion, I can check if a client is connected or not: with this one, not longer.- with this one, when the client disconnect, I close the master pty and open it again, with the risk to have a different client pty.So ... I prefer the old one (more standard accross various plateform) and I think the solution used to solve the issue solved by the new one is maybe not the correct one.send from my mobile.____________________FrancoisLe 18 mai 2024 à 18:50, tbodt ***@***.***> a écrit :﻿
Do you run into major issues with the new behavior?

cgull:
I think I've fixed this in #2449.  Sorry for dropping the ball on it.  I'm curious, though, what sort of application would you be running in iSH that needs this?  The things I can think of include embedded applications or applications needing to connect to or emulate physical devices, none of which seem like something you'd want to do in an iOS app.

PR #2451, opened 2024-09-04
Promote hostname override to a friendly preference 

saagarjha:
Users really hate that we put "localhost" in their prompt as a result of iOS 16 anti-fingerprinting obfuscating the device name. We already have a mechanism for overriding this that we use for App Store screenshots, so we might as well make it official and expose it for those that care.

The logic for what the actual uname hostname is as follows:

* If the user sets the preference, use that.
* On iOS <16, if the preference is unset, use what real uname returns.
* On iOS 16+, use UIDevice.name. This at least changes "localhost" to something like "iPad".

### DIFF
--- a/app/AppDelegate.m
+++ b/app/AppDelegate.m
@@ -284,7 +284,10 @@ void NetworkReachabilityCallback(SCNetworkReachabilityRef target, SCNetworkReach
     proc_ish_version = strdup(ishVersion.UTF8String);
     // this defaults key is set when taking app store screenshots
     extern const char *uname_hostname_override;
-    NSString *hostnameOverride = [NSUserDefaults.standardUserDefaults stringForKey:@"hostnameOverride"];
+    NSString *hostnameOverride = UserPreferences.shared._hostnameOverride;
+    if (@available(iOS 16.0, *)) { // Hostname obfuscation is in effect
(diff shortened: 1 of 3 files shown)


emkey1:
This is great, but it would also be good to support the system call that sets the hostname.  It's trivial.

saagarjha:
Making it do something would be trivial, yes, but since it doesn't actually change the hostname we figure that returning EPERM is the right thing to do. The user preference key is pretty clear about what it does and this is purely being promoted because *we* use it and might as well put it in /proc/ish/defaults rather than have people poke around in /proc/ish/.defaults for it. With that said, if you have a usecase that you think would be better fulfilled by changing sethostname, please let us know.

emkey1:
It's not a huge deal either way, but people with a Linux/Unix background expect the hostname command to be able to set the hostname as well as showing it, so it's confusing when it fails at what is essentially a root prompt.

It doesn't look like sethostname() is defined  in POSIX, but it's also present in pretty much every Unix like operating system from what I can tell.  Again, not a big deal but something to think about.  I guess in part it depends on how closely you want to adhere to Unix/Linux conventions (when feasible).

tbodt:
Why don't we have sethostname set the user defaults key

saagarjha:
Because the personal test I have in mind for whether this should actually be supported is whether you can set the hostname and then SSH in from another machine with the new hostname. If you just want to change PS1, well, this is for you.

PR #2594, opened 2025-06-29
Always focus on the window when there is no selection

parport0:
Selection.rangeCount can return 1 when there is no selection, it is enough for a selection to have been present in the past or for the user to have clicked on the page.

See MDN: https://developer.mozilla.org/en-US/docs/Web/API/Selection/rangeCount

Use Selection.isCollapsed to be entirely sure there is a selection before ignoring the click.

This change fixes the issue ish-app/ish#2519, where focus to the terminal is lost and can not be regained, forcing the user to restart ish.

Tested on iPad Pro 13-inch (M4) with iPadOS 26 Beta 2 and Stage Manager.


hakornyutt-create:
87f7b1d8a60b6c4b53f5da526e805e2b327b6bdd

PR #2603, opened 2025-07-19
feat: Asbestos JIT 引擎极致性能重构

neipor:
对 iSH 核心 Asbestos JIT 引擎进行全面重构，实现极致性能优化


saagarjha:
While we appreciate the contribution, I don’t think we can accept this change as-is, since it has too many unrelated modifications. If you’d like to revisit this please split up your changes into smaller, isolated diffs and clearly explain what the performance improvements are intended to be (plus how you benchmarked it). I skimmed over the PR and I remain unconvinced these would help in real-world use. If you need help designing benchmarks I’d be happy to help you out.

neipor:
sorry my fault :(

PR #2646, opened 2025-10-24
Add show op for /proc/pid/mem

jhapradip:
lseek on /proc/pid/mem calls the 'show' op to set up metadata for the file. Currently, this API is not defined for this file and this leads to a crash when gdb launches any program. I have added a default implementation as no special handling is required. Other sequences such as (p)read/(p)write work as expected.

### DIFF
--- a/fs/proc/pid.c
+++ b/fs/proc/pid.c
@@ -172,6 +172,16 @@ out_free_task:
     return err;
 }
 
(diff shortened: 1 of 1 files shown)


ztfzxxcm28-ux (approved):

ztfzxxcm28-ux (commented):
ڤ

PR #2708, opened 2026-03-21
Forward touch scroll events as terminal input

### LINKED ISSUE #2375: Allow for mouse wheel scrolling to be converted to up/down arrows for tmux/screen/vim copy mode
It would be really nice to have a way to use external mouse scroll wheel or trackpad scrolling to simulate up/down (maybe even left/right) arrow key presses. Maybe a toggle in the settings window to switch between normal window scrolling behavior and simulated key presses. 

Thanks!

### LINKED ISSUE #2537: No Mouse Wheel Event, plz make it available to 3d party programs
Hi ! I use tintin++ with ish, run quite well (sometimes input freez) anyway. 
Could you add mose wheel event support? I use very often external mouse but in the therminal  make the buffer scroll but doen't raise in tintin++. Could you make it available?

splch:
Dispatches synthetic `WheelEvent`s into hterm's existing `onMouse_` pipeline so that touch scroll gestures reach programs that listen for them.

**When mouse reporting is active** (e.g. tmux `set -g mouse on`), hterm's `VT.onTerminalMouse_` generates the appropriate escape sequences. **When on the alternate screen** (less, vim, man), hterm's built-in alternate scroll mode (DECSET 1007) sends arrow keys. **On the primary screen with no mouse reporting**, the handler is a no-op and existing scrollback behavior is unchanged.

### Changes

`TerminalView.m`: compute scroll delta in `scrollViewDidScroll:` and forward to JS
`term.js`: add `handleScrollDelta` export that dispatches `WheelEvent`s to `scrollPort_.screen_`

### Test plan

- [ ] tmux with `set -g mouse on`: touch scroll enters copy-mode and scrolls
- [ ] `less` / `man`: touch scroll moves through content
- [ ] Primary screen shell prompt: scrollback works as before

Fixes #2375
Fixes #2537

### DIFF
--- a/app/TerminalView.m
+++ b/app/TerminalView.m
@@ -272,9 +272,11 @@ static NSString *const HANDLERS[] = {@"syncFocus", @"focus", @"newScrollHeight",
 
 - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
     CGFloat newY = scrollView.contentOffset.y;
+    [self.terminal.webView evaluateJavaScript:[NSString stringWithFormat:@"exports.newScrollTop(%f)", newY] completionHandler:nil];
     CGFloat delta = newY - _prevScrollY;
     _prevScrollY = newY;
-    [self.terminal.webView evaluateJavaScript:[NSString stringWithFormat:@"exports.handleScroll(%f, %f)", newY, delta] completionHandler:nil];
+    if (delta != 0)
+        [self.terminal.webView evaluateJavaScript:[NSString stringWithFormat:@"exports.handleScrollDelta(%f)", delta] completionHandler:nil];
 }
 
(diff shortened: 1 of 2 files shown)


tbodt:
This feels too independent from hterm. Surely hterm already supports passing scroll events? iSH should reuse that as much as possible

splch:
> This feels too independent from hterm. Surely hterm already supports passing scroll events? iSH should reuse that as much as possible

Great point i'm looking into this right now

splch:
Note: Synthetic WheelEvents use default clientX/clientY (0, 0), so mouse wheel reports always map to terminal coordinates (1, 1). This works fine for scrolling in tmux, vim, less, etc., but programs that need accurate scroll position won't get it

PR #2723, opened 2026-04-27
Fuse aarch64 register moves

siraben:
## Summary

Fuse 32-bit register-to-register `mov` on aarch64 into one direct gadget instead of emitting separate load and store gadgets.

Changes:
- Add direct aarch64 `mov32_reg_reg` gadgets for the general register matrix.
- Route 32-bit register-to-register MOV generation through the direct gadget table on aarch64.
- Keep non-aarch64 and non-register MOV generation on the existing load/store path.

## Results

Measured with the benchmark/profiling branch now kept separately at `siraben/ish:bench/hot-system-paths-bench-profile`. Release build, 7 runs each:

| benchmark | baseline avg | optimized avg | change |
| --- | ---: | ---: | ---: |
| python_startup | 0.157s | 0.151s | -3.8% |
| python_compute | 8.179s | 7.524s | -8.0% |
| python_imports | 2.039s | 1.846s | -9.5% |
| bash_control | 36.543s | 34.936s | -4.4% |
| shell_pipeline | 0.600s | 0.561s | -6.5% |

Instrumentation also showed translated block code words dropping for representative workloads:
- bash control: `144074 -> 142484`
- python compute: `896963 -> 886056`

## Validation

- `ninja -C build`

### DIFF
--- a/asbestos/gadgets-aarch64/math.S
+++ b/asbestos/gadgets-aarch64/math.S
@@ -13,6 +13,42 @@
     strh _tmp, [_cpu, #CPU_gs]
     gret
 
+.macro mov32_reg_reg dst_name, dst, src_name, src
+    .gadget mov32_\dst_name\()_\src_name
+        mov \dst, \src
+        gret
+.endm
+
+.macro mov32_reg_reg_row dst_name, dst
+    mov32_reg_reg \dst_name, \dst, reg_a, eax
+    mov32_reg_reg \dst_name, \dst, reg_c, ecx
+    mov32_reg_reg \dst_name, \dst, reg_d, edx
+    mov32_reg_reg \dst_name, \dst, reg_b, ebx
+    mov32_reg_reg \dst_name, \dst, reg_sp, esp
+    mov32_reg_reg \dst_name, \dst, reg_bp, ebp
+    mov32_reg_reg \dst_name, \dst, reg_si, esi
+    mov32_reg_reg \dst_name, \dst, reg_di, edi
+.endm
+
+mov32_reg_reg_row reg_a, eax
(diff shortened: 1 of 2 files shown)


asbestos/gen.c
```diff
@@ -212,6 +213,45 @@ static inline bool gen_op(struct gen_state *state, gadget_t *gadgets, enum arg a
         GEN(state->orig_ip | state->orig_ip_extra);
     return true;
 }
+
+static inline enum arg gen_resolve_arg(enum arg arg, struct modrm *modrm, uint64_t *imm, dword_t addr_offset) {
```
saagarjha, line 217: This is basically the same code as what is right above it, can we avoid duplicating it? We aren't even using its full functionality; I think we should probably just fastpath the things we care about

asbestos/gadgets-aarch64/math.S
```diff
@@ -13,6 +13,42 @@
...
+.endm
+
+mov32_reg_reg_row reg_a, eax
+mov32_reg_reg_row reg_c, ecx
+mov32_reg_reg_row reg_d, edx
+mov32_reg_reg_row reg_b, ebx
+mov32_reg_reg_row reg_sp, esp
+mov32_reg_reg_row reg_bp, ebp
+mov32_reg_reg_row reg_si, esi
+mov32_reg_reg_row reg_di, edi
```
saagarjha, line 40: I think we have an each_reg macro for this or something like that

saagarjha (changes requested):
Nice work! I measured a consistent 2-5% win across various workloads which seems in line with what you were seeing. A couple of comments though, and also if you could make an x86-64 path for this optimization too that would be much appreciated.

PR #2724, opened 2026-04-27
Create standard devices for command line runs

siraben:
## Summary

Create standard `/dev` nodes during command-line emulator startup.

The iOS app startup already creates `/dev/ptmx`, but the CLI path only mounted
`devpts` at `/dev/pts`. Static i386 `tmux` could start and parse commands, but
failed when creating the first pane because `forkpty` could not open the pty
master:

```text
create window failed: fork failed: No such file or directory
```

This moves the existing standard device-node setup into shared kernel init code
and calls it from both app and CLI startup. The CLI path now gets `/dev/ptmx`
before mounting `devpts`, without duplicating the app's `/dev` setup list.

## Validation

- `ninja -C build`
- Removed `/dev/ptmx` from the Alpine test filesystem, then confirmed the
  rebuilt CLI emulator recreated it on startup.
- Ran static i386 `tmux` in Alpine:
  - `tmux -V`
  - `tmux -h`
  - `tmux -L final-smoke -f /dev/null new-session -d /bin/true`

### DIFF
--- a/kernel/init.c
+++ b/kernel/init.c
@@ -3,9 +3,11 @@
 #include <sys/stat.h>
 #include "fs/devices.h"
 #include "fs/fd.h"
+#include "fs/path.h"
 #include "fs/real.h"
 #include "fs/tty.h"
 #include "kernel/calls.h"
+#include "kernel/fs.h"
 #include "kernel/init.h"
 #include "kernel/personality.h"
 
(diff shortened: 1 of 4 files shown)


kernel/init.c
```diff
@@ -125,6 +127,30 @@ void set_console_device(int major, int minor) {
     console_minor = minor;
 }
 
+static void create_device_node(const char *path, int major, int minor) {
```
tbodt: I don't really think factoring out this into a function improves the code, nor does the loop from 1 to 7, could you undo that and copy the code as is

kernel/init.c
```diff
@@ -125,6 +127,30 @@ void set_console_device(int major, int minor) {
...
+    generic_mknodat(AT_PWD, path, S_IFCHR|0666, dev_make(major, minor));
+}
+
+void create_standard_devices(void) {
```
tbodt: a better name is create_some_device_nodes. who says they're standard?

tbodt:
theoretically the correct solution is to create an implementation of devtmpfs instead of creating device nodes like this. (i don't think I knew about devtmpfs at the time)

PR #2744, opened 2026-05-27
signal: remove SIGURG from default-ignore list

badoriie:
## Problem

`SIGURG` is currently in iSH's default-ignore list in `kernel/signal.c`:

```c
case SIGURG_: case SIGCONT_: case SIGCHLD_:
case SIGIO_: case SIGWINCH_:
    return SIGNAL_IGNORE;
```

**Go 1.14+** uses `SIGURG` to implement asynchronous goroutine preemption (`runtime/signal_unix.go`). When a Go program calls `signal.Notify` or the runtime installs its own `SIGURG` handler via `runtime_sigInstall`, iSH's default-ignore classification prevents proper handler registration.

The result is a hard crash on startup:
```
fatal: bad g in signal handler
```

The workaround is to set `GODEBUG=asyncpreemptoff=1` before launch, which disables Go's async preemption entirely — a significant runtime regression.

## Fix

Remove `SIGURG_` from the ignore list. Its POSIX default disposition is already "ignored" for processes that haven't installed a handler, so removing it from this list has no effect on programs that don't use `SIGURG`. It only allows programs like Go to successfully install their own handler.

```c
// Before
case SIGURG_: case SIGCONT_: case SIGCHLD_:

// After
case SIGCONT_: case SIGCHLD_:
```

## Impact

- Go 1.14+ programs start without `GODEBUG=asyncpreemptoff=1`
- No behavioural change for programs that don't use `SIGURG`
- Tested with [Conduit](https://github.com/Psiphon-Inc/conduit) on iSH (Alpine Linux, iPhone)

### DIFF
--- a/kernel/signal.c
+++ b/kernel/signal.c
@@ -38,7 +38,7 @@ static int signal_action(struct sighand *sighand, int sig) {
     }
 
     switch (sig) {
-        case SIGURG_: case SIGCONT_: case SIGCHLD_:
+        case SIGCONT_: case SIGCHLD_:
         case SIGIO_: case SIGWINCH_:
             return SIGNAL_IGNORE;
 


emkey1:
I tested this one hoping to corroborate it, but the measurements point the other way, so passing them along rather than sitting on them.

**Master already behaves correctly, and this patch changes it to something Linux does not do.**

`signal_action()` checks the installed disposition *before* reaching that switch:

```c
static int signal_action(struct sighand *sighand, int sig) {
    if (signal_is_blockable(sig)) {
        struct sigaction_ *action = &sighand->action[sig];
        if (action->handler == SIG_IGN_)
            return SIGNAL_IGNORE;
        if (action->handler != SIG_DFL_)
            return SIGNAL_CALL_HANDLER;    // installed handler wins here
    }
    switch (sig) {
        case SIGURG_: ... return SIGNAL_IGNORE;   // only reached for SIG_DFL
```

So the list only sets the **default** action. A process that installs a SIGURG handler — which is what the Go runtime does for async preemption — returns at `SIGNAL_CALL_HANDLER` and never reaches the switch.

And per `signal(7)`, SIGURG's default action **is** Ign, so the entry is correct as written.

Measured with one static i386 binary run three ways — it installs a handler and raises SIGURG, then forks a child that resets to `SIG_DFL` and raises it again:

| | installed handler runs | default-disposition SIGURG |
|---|---|---|
| real Linux | yes | ignored, child exits 42 |
| upstream master | yes | ignored, child exits 42 |
| master + this PR | yes | **child killed by signal 23** |

Master matches Linux exactly on both. With the patch, a default-disposition SIGURG kills the process, which no Linux program expects — and it does not change the installed-handler case at all, which is the one Go depends on.

If Go's preemption really is misbehaving under iSH, the cause looks like it is somewhere else — signal delivery to a blocked thread, or restart semantics, rather than the disposition table. Happy to help chase it with the same setup if you have a Go reproducer; I have a real-Intel Linux box wired up for exactly this kind of A/B.

PR #2754, opened 2026-06-20
Add option to use the extra space on iphones with bevels.

simlay:
I felt like there's a bunch of unused space so I tweaked some of the padding and option to maximize the screen.

* The webview and scrollview are sharing the screen so there's tweaking on the extraWidth and extraHeight to force the webview to overlap on the scrollview.
* The Pad at the bottom of the screen I made optional via the "Maximize Screen Space".  Doing this correctly really requires "Hide with external keyboard" option under "extra keys" as the webview will render under the extra keys bar.
* The padding in the webview being dynamic via `setScreenPaddingSize` also reduces the padding when there's an external keyboard.

The removal of random white space is because my vim config removes it. I can revert those changes if you'd like.

Full disclosure, I had qwen3.6 help me via a local llm to suggest the changes as I don't know objective-c well.


saagarjha:
Can you provide a screenshot of what this is intended to do? It's not quite clear what your goal is with this change.

simlay:
I'm still coming back to this pr. I published it so I could feel like I made progress but left it as draft because it's not up to my standards. I'll be adding some before/after images and why the changes are relevant.

PR #2763, opened 2026-07-09
Add and fix CI workflow to include emulator tests

devops2626:


saagarjha:
Closing for similar reasons as #2777

PR #2774, opened 2026-07-27
Return EINVAL from getdents when the buffer is too small

emkey1:
If the caller's buffer cannot hold even the first entry, the loop just breaks and `sys_getdents_common` returns 0. Every caller reads 0 as end of directory, so a `readdir()` with a small buffer quietly reports the directory as **empty** rather than failing.

Linux returns `EINVAL` for this case. This rewinds to the entry that did not fit and returns that instead. Only the first iteration can be affected — once something has been written, returning a short count is correct, so the existing behavior for partial fills is unchanged.

### Testing

Repro: `getdents64` on `/` with an 8-byte buffer.

| | result |
|---|---|
| Linux (x86 Ubuntu) | `-1 (Invalid argument)` |
| master | `0` — reads as an empty directory |
| this branch | `-1 (Invalid argument)` |

Also ran a shell smoke test over an Alpine rootfs (`ls`, `find`, directory walks with normal-sized buffers, `/proc` reads, fork/exec churn, redirects, pipes) with no change in behavior.

### DIFF
--- a/fs/dir.c
+++ b/fs/dir.c
@@ -92,8 +92,15 @@ int_t sys_getdents_common(fd_t f, addr_t dirents, dword_t count,
             printed++;
         }
 
-        if (reclen > count)
+        if (reclen > count) {
+            if (count == orig_count) {
+                // If the buffer isn't large enough to read even a single dirent,
+                // reset the fd and return _EINVAL.
+                fd_seekdir(fd, ptr);
+                return _EINVAL;
+            }
             break;
+        }
         if (user_write(dirents, dirent_data, reclen))
             return _EFAULT;
         dirents += reclen;


fs/dir.c
```diff
@@ -92,8 +92,16 @@ int_t sys_getdents_common(fd_t f, addr_t dirents, dword_t count,
             printed++;
         }
 
-        if (reclen > count)
+        if (reclen > count) {
+            // Linux returns EINVAL when the buffer cannot hold even the first
+            // entry. Returning 0 here instead would read as end-of-directory,
+            // so a readdir() with a small buffer sees an empty directory.
```
saagarjha: The default behavior for iSH is to match what Linux does, so there is no need to call attention to it. In fact it is generally a good idea to point out places where behavior (intentionally) diverges.

```suggestion
            // If the buffer isn't large enough to read even a single dirent,
            // reset the fd and return _EINVAL.
```

saagarjha (changes requested):
This looks semantically correct but I am not entirely sure how you ran into this behavior; I cannot imagine that there are many programs that pass in a buffer this small. Some tips for the comment style though

emkey1:
Comment reworded as suggested, and noted for future patches — flag divergence, not conformance. Pushed.

On how I ran into it: honestly, not from a program failing. I was running a differential conformance suite against real Linux — same static i386 binary on both, compare the results — and this fell out of the `getdents` cases. So it is a conformance fix rather than a bug report, and I do not have a real-world consumer to point at. I share your instinct that few programs pass a buffer that small.

The one thing that made it seem worth sending is the failure mode: returning 0 is indistinguishable from end-of-directory, so a caller with an undersized buffer sees an empty directory rather than an error. That is the sort of thing that would be very hard to diagnose from the guest side if it ever did bite. But if the bar is an observed problem, that is a fair call and I am happy for you to close it.

DLeylandOY (approved):

saagarjha:
I am going to merge this as it is a real behavior divergence from Linux but in the future please keep in mind that we have limited time and would prefer to review changes that increase compatibility instead of fixing rarer edge cases

PR #2800, opened 2026-09-06
Several Potential Bug Fixes

owenallen-sudo:
I've got several bug fixes that would likely work and would have almost no impact if they don't. The PR includes fixes for race conditions, integer overflows, buffer overflows, null pointer dereferences and more. There are also code cleanups such as the duplicate #include "fs/devices.h" in app/terminal.m removed. My fixes are in app/terminal.m, app/terminal/term.js, platform/darwin.c, asbestos/gadgets-aarch64/, app/LinuxTTY.c, app/LinuxPTY.c, platform/platform.h and more. The only downside I have found with my patches is that scrolling is briefly locked when doing clear command but that is just one of my patches. Not all of the fixes are actual bugs or cleanups, some are just fixing vulnerabilities incase there is a vulnerability in the Apple sandbox. Many however are actual bugs. I can't verify with certainty my patches work and produce no downsides though.

### DIFF
--- a/.github/workflows/autolabel.yml
+++ b/.github/workflows/autolabel.yml
@@ -5,14 +5,19 @@ on:
 jobs:
   label:
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
     steps:
-      - uses: satackey/action-js-inline@bf6fcaf35de1ed03bcfd25a0a8b1fa4c551ec908
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/github-script@v7
         with:
           script: |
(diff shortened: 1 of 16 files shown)


DLeylandOY (approved):

RepoKan (approved):

saagarjha:
Hi, and thanks for your interest in contributing to iSH! While we appreciate your candor about your PR, we really cannot accept bug fixes that "would likely work", especially as they span multiple parts of the codebase. If you're still like to submit these changes, please identify specific issues (instead of a big group together) and keep your PR scoped to the relevant code only to ease review for us. Thanks!

